PR #1, opened 2015-11-21
initial commit

sarahgripshover:


util_function_groups/column_operations.R
```diff
@@ -0,0 +1,137 @@
...
+    "race" = list(
+        "datatype" = "character",
+        "accepted_values" = c("White", "Black", "Latino", "Asian", "Other", "Unknown"),
+        "column_required" = FALSE,
+        "blanks_allowed" = TRUE
+    ),
+    "ddowell" = list(
+        "datatype" = "numeric",
+        "accepted_range" = c(1, 7),
+        "blanks_allowed" = TRUE
```
cmacrander: This is a little nervous-making to me... what's a blank? `NA` in R, `null` in JSON? We should write down somewhere in this file all the flavors of blank and how they're treated... the empty string being an especially slippery one.

util_function_groups/column_operations.R
```diff
@@ -0,0 +1,137 @@
...
+                       blanks_allowed = validate_blanks_allowed(column_attributes,values),
+                       accepted_values = validate_accepted_values(column_attributes,values),
+                       accepted_range = validate_accepted_range(column_attributes,values),
+                       datatype = validate_datatype(column_attributes,values)
+                )
+            }
+        }
+        validate_required(column_attributes, names(df))
+        # Finally, check that the column exists, if it's required
+        # If it was, throw a warning.
```
cmacrander: Should this comment be above the function call on line 83? Right now it looks like the comment is talking about code which is missing...

util_function_groups/column_operations.R
sarahgripshover: hmm, I didn't think about blanks in json, just R. And I used the util.is_blank() function to find the blanks, which would not pick out a `null` value:
function(x){
  is.na(x) | grepl("^[ \t]*$", x)
}
It seems like making this whole thing compatible with json output is a big source of value, so probably worth it to add `null` to the list of blank values. Honestly, could we do that by adding `null` to util.is_blank, i.e.,
  is.na(x) | grepl("^[ \t]*$", x) | x %in% "null"
I can't imagine a case where we'd have a value in a cell that's literally the string "null" and we wouldn't want to just call it blank.

util_function_groups/column_operations.R
sarahgripshover: fixed this!

PR #2, opened 2015-11-23
New math functions

daveponet:
I'm going to merge these in because they're tiny and -- I hope -- uncontroversial. Please feel free to nit!

@cmacrander @dkgreene @sarahgripshover

### DIFF
--- a/util_function_groups/math.R
+++ b/util_function_groups/math.R
@@ -1,13 +1,18 @@
 
 
-# Basic Data Transformation
 util.z_score <- function(x){
+  # calculate the z-score of vector or for each vector in a data.frame
   if(class(x) %in% "data.frame"){
-    do.call(cbind,lapply(x, util.z_score))
+    util.apply_columns(x,util.z_score)
   }
   else{ ( x - mean(x,na.rm=T) ) / sd(x,na.rm=T) }
 }
 
+util.row_means <- function(x){
+  # like base::rowMeans, but returns self if vector
+  if(is.vector(x)) return(x)
+  rowMeans(x, na.rm = TRUE)
+}
 
 # round all numeric columns
 util.round_df <- function(DF, digits=2, show_caution=TRUE){
--- a/util_function_groups/math.R
+++ b/util_function_groups/math.R
@@ -9,9 +9,17 @@ util.z_score <- function(x){
 }
 
 util.row_means <- function(x){
-  # like base::rowMeans, but returns self if vector
+  # like base::rowMeans(x, na.rm=TRUE) 
+  # but it returns self if vector instead of issuing an error
   if(is.vector(x)) return(x)
   rowMeans(x, na.rm = TRUE)
+}
+
+util.row_sums <- function(x){
+  # like base::rowSums(x, na.rm=TRUE) 
+  # but it returns self if vector instead of issuing an error
+  if(is.vector(x)) return(x)
+  rowSums(x, na.rm = TRUE)
 }
 
 # round all numeric columns


cmacrander:
How about a comment on why the forced na.rm?

sarahgripshover:
only nit is that it seems a bit weird to me to have something below the
return line, but maybe that's a thing and I'm not aware of it?

On Mon, Nov 23, 2015 at 3:52 PM, Chris <notifications@github.com> wrote:

> How about a comment on why the forced na.rm?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PERTS/gymnast/pull/2#issuecomment-159107085>.
>



-- 
Sarah Gripshover, PhD
Program Manager
PERTS
Department of Psychology
Stanford University

daveponet:
@sarahgripshover it seems like any other version of this would be a lot bulkier. I don't think having a return above other code is weird, especially not in a function with only two lines -- both of which are returns.

cmacrander:
I vote for not mixing explicit and implicit returns in one function, so here just adding `return()` to the last line of each.

cmacrander:
You could also use `ifelse()`:

```
return(ifelse(is.vector(x), x, rowSums(x, na.rm = TRUE))
```

daveponet:
ifelse for the win!

daveponet:
Actually, `ifelse` doesn't work because it assumes it assumes the test variable is vectorized if the yes value is vectorized. 

```r

util.row_means <- function(x){
  # like base::rowMeans(x, na.rm=TRUE) 
  # but it returns self if vector instead of issuing an error
  if(is.vector(x)) return(x)
  return(rowMeans(x, na.rm = TRUE))
}

util.row_means_ifelse <- function(x){
    # like base::rowMeans(x, na.rm=TRUE) 
    # but it returns self if vector instead of issuing an error
    return(ifelse(is.vector(x), x, rowMeans(x, na.rm = TRUE)))
}

x <- c(1,2,3,NA)
util.row_means(x)
# returns  1  2  3 NA
util.row_means_ifelse(x)
# 1

```

sarahgripshover:
bleh that's annoying.

Does

if(is.vector(x)){
    return(x)
} else{
return(rowMeans(x, na.rm = TRUE))
}

make you want to hurl? Seems simple to me but I also think we're probably
nitpicking this way too much & you could just do it the way you had
originally and it would be fine


On Wed, Nov 25, 2015 at 7:34 PM, daveponet <notifications@github.com> wrote:

> Actually, ifelse doesn't work because it assumes it assumes the test
> variable is vectorized if the yes value is vectorized.
>
> util.row_means <- function(x){
>   # like base::rowMeans(x, na.rm=TRUE)
>   # but it returns self if vector instead of issuing an error
>   if(is.vector(x)) return(x)
>   return(rowMeans(x, na.rm = TRUE))
> }
> util.row_means_ifelse <- function(x){
>     # like base::rowMeans(x, na.rm=TRUE)
>     # but it returns self if vector instead of issuing an error
>     return(ifelse(is.vector(x), x, rowMeans(x, na.rm = TRUE)))
> }
> x <- c(1,2,3,NA)
> util.row_means(x)# returns  1  2  3 NA
> util.row_means_ifelse(x)# 1
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PERTS/gymnast/pull/2#issuecomment-159793926>.
>



-- 
Sarah Gripshover, PhD
Program Manager
PERTS
Department of Psychology
Stanford University

daveponet:
I ended up making it the way I had it originally, but with an explicit return statement the end. So it's like your code without the `else`.

PR #3, opened 2015-11-24
Var renaming

dkgreene:
This is the function Chris and I developed during the code retreat.


util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
...
+################################ testing for rename_columns  ################################################
+
+######################## TEST 1: handling forbidden names
+
+
+test_forbidden_names <- function() {
+  
+  my_mapping <- list(
+    old_name = c("foo", "bar", "baz","blerp"),
+    new_name = c("newfoo", "newbar", "newbaz", NA)
```
cmacrander: Pretty sure this should be `"NA"`, not `NA`.

util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
...
+  
+  tryCatch(
+    expr = {
+      rename_columns(d,my_mapping)
+    },
+    error = function(e) {
+      success <<- TRUE
+    }
+  )
+  return(success)
```
cmacrander: Let's add a todo in the comments about abstracting this "assert that the function raises an error" code.

util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
...
+  starting_w_digit_pattern <- "^\\.?[0-9]"
+  naughty_words <- c("if", "else", "repeat", "while", "function", "for",
+                     "in", "next", "break", "TRUE", "FALSE", "NULL", "Inf",
+                     "NaN", "NA", "NA_integer_", "NA_real_", "NA_complex_", "NA_character_", NA)
+  
+  # if the new names in the mapping match any evil patterns, then complain
+  if( any(mapping$new_name %in% naughty_words) |
+      any(grepl(whitespace_pattern, mapping$new_name)) |
+      any(grepl(starting_w_digit_pattern, mapping$new_name))) {
+    stop("Invalid new names are present among: ",paste(mapping$new_name, collapse =" "))
```
cmacrander: @dkgreene messy whitespace here and elsewhere, please go over the file to check.

For example, this line should be:

`   stop("Invalid new names are present among: ", paste(mapping$new_name, collapse = " "))`

util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
 source("~/Sites/gymnast/util.R", chdir = FALSE)
 source("~/Sites/gymnast/util_qualtrics_cleaning.R", chdir = FALSE)
 
+
+############################### rename_columns function  #################################################
+
+rename_columns <- function(data, mapping) {
```
daveponet: This should be renamed `util.rename_columns()`

util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
...
+    newfoo = c(1,2,3,4,5)
+  )
+  
+  new_d <- rename_columns(d,my_mapping)
+  
+  return(identical(new_d, expected_output))
+  
+}
+
+test_extra_mapping_columns()
```
daveponet: I'd like to use a regular naming pattern for the unit tests so we can clear them all out the namespace once util loads. How about `gymnast_test.[testname]`?

util_function_groups/column_operations.R
```diff
@@ -3,6 +3,219 @@
 source("~/Sites/gymnast/util.R", chdir = FALSE)
```
daveponet: We can't source util from the subfiles. That would get us into a recursive loop.

util_function_groups/column_operations.R
cmacrander: since this would be a gymnast-internal thing and super careful scoping isn't so important, I think just the prefix `test_` is fine.

But automatically clearing them from the defined variables as gymnast loads would be sweet. Although I think there are even better namespacing options out there somewhere...

daveponet: @cmacrander how would it be a "gymnast-internal thing" unless we clear it out of the general scope at the end?

cmacrander: Sorry for being confusing: I'm assuming we _will_ clear it out at the end, b/c that's such a good idea.

daveponet: @cmacrander the reason I thought `test_` wasn't good enough was because I was afraid of clearing all objects from the global scope that start with that name pattern.

cmacrander: Oh, that's right, b/c you don't know where in their code someone else will source gymnast. Well then I think better than relying on a fancy name is wrapping it in a function:

```
(function () {
    tests <- c(
        function () print("it works!"),
        function () print("so fun!")
    )
    for (t in tests) {
        t()
    }
})()
```

util_function_groups/column_operations.R
dkgreene: It could be either - both are on the naughty words list above

cmacrander: Oh, I see. Withdrawn!

PR #5, opened 2015-12-01
Util consolidation

daveponet:
@cmacrander code review requested! 

util subfiles consolidated into util.R.
apply_columns accepts ellipsis;
to_stripped_characters removed (can just apply_columns on
strip_special_characters instead);
util.round_df uses improved apply_columns


cmacrander:
Some comments to add and dead code to kill, otherwise :moneybag:

daveponet:
Thanks, @cmacrander! All suggested adopted.

PR #6, opened 2015-12-10
Recode warning

sarahgripshover:
@daveponet see if you like the warning I added to your recode function

### DIFF
--- a/util_function_groups/value_replacement.R
+++ b/util_function_groups/value_replacement.R
@@ -1,6 +1,35 @@
 
 # replace appearances of originals with replacements inside the vector
 util.recode <- function(vector, originals, replacements){
+  # check if any of the replacement values also appear in the originals.
+  if(any(replacements %in% originals)){
+    # if they do:
+    # check whether any shared values appear earlier in the list of
+    # replacements than in the originals. If they do,
+    # this will cause desired values to be over-written. For example, if
+    # originals == c("a","b") and replacements == c("b", "c"), then you
+    # probably want the vector c("a", "b", "a") to be recoded as c("b", "c",
+    # "b"). But what you will actually get is c("c", "c", "c"), because the
+    # "a"s get replaced with "b"s before the b's get replaced with c's. So warn
+    # if this is happening, otherwise it's super-annoying to track down!
+    # find the shared values
+    shared_values <- replacements[replacements %in% originals] %>% unique
+    # find their positions
+    first_shared_replacements <- which(replacements %in% shared_values) %>% min
+    first_shared_originals <- which(originals %in% shared_values) %>% min
+    # warn if any shared values appear earlier in replacements than originals
+    if(first_shared_replacements < first_shared_originals){
+        util.warn("In util.recode()," %+%
+            "you have some values in originals that also " %+%
+            "appear in replacements. Some of these shared values may " %+%
+            "cause recoded values to be over-written. If you don't " %+%
+            "want this to happen, please check your recoded output " %+%
+            "to make sure no desired values were over-written. " %+%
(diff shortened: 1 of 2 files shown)


util_function_groups/value_replacement.R
```diff
@@ -1,6 +1,35 @@
 
 # replace appearances of originals with replacements inside the vector
 util.recode <- function(vector, originals, replacements){
+  # check if any of the replacement values also appear in the originals.
+  if(any(replacements %in% originals)){
```
daveponet: @sarahgripshover, great catch. This is indeed a very surprising behavior.

However, I'm confused. Do you think this is the desired behavior? I.e., do you think that the vector in your example _should_ turn into all "c"s? It seems not only surprising, but also just downright wrong. If you said you wanted "a" turned into "b", then surely the function is broken if it returns "c" for "a". Right?

Given that this function is -- in my estimation -- broken. The right remedy is not to throw a warning. It's to fix the function. I think the best ways to deal with that would be to:

* Fix it yourself.
* Comment on my initial commit that it's messed up so I can fix it (or whoever was responsible for submitting broken code).
* Use github to start an issue (this is probably the most appropriate).


Anyway, here's my recommended new version of this function, along with a test that would fail using the original version of the function. 

```r
util.recode <- function(vector, originals, replacements){
    # replace appearances of "originals" with "replacements"
    if(length(originals) != length(replacements)){
        "Original and replacement should have equal lenght" %>%
            util.warn()
    }
    if(any(duplicated(originals))){
        "Originals should not include duplicate values" %>%
            util.warn()
    }
    
    new_vec <- vector
    for(v in originals){
        new_vec[vector == v] <- replacements[originals == v]
    }
    new_vec
}

gymnast_test__util.recode <- function(){
    # basic test
    test_name <- "basic replacement"
    vector <- c("a","b","fefe","c")
    x <- c("a","b","c")
    y <- c("A","B","C")
    expectation <- c("A","B","fefe","C")
    if( !identical(expectation,util.recode(vector,x,y)) ){
        util.warn("recode fails: " %+% test_name)
    }
    
    test_name <- "multiple replacement includes originals"
    vector <- c("a","b","fefe","c")
    x <- c("a","b","c")
    y <- c("b","c","d")
    expectation <- c("b","c","fefe","d")
    if( !identical(expectation,util.recode(vector,x,y))){
        util.warn("recode fails: " %+% test_name)
    }
}
```

Finally, I just want to note that you're making changes in a branch that's very much out of date (by 8 days). In fact, the file you're working on no longer exists on master. I made a separate pull request with the code above: 

https://github.com/PERTS/gymnast/pull/7/files

daveponet:
@sarahgripshover, 
Given that the main substantive change was handled by #7, maybe we should just delete this one? Maybe just add the comment you made in `util_qualtrics_cleaning.R` directly to master... might be easier than resolving conflicts (as you'd need to in order to merge this request).

sarahgripshover:
Yeah, I'll delete this branch and add the comment in master.

On Thu, Dec 10, 2015 at 6:36 PM, Dave Paunesku <notifications@github.com>
wrote:

> @sarahgripshover <https://github.com/sarahgripshover>,
> Given that the main substantive change was handled by #7
> <https://github.com/PERTS/gymnast/pull/7>, maybe we should just delete
> this one? Maybe just add the comment you made in util_qualtrics_cleaning.R
> directly to master... might be easier than resolving conflicts (as you'd
> need to in order to merge this request).
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/PERTS/gymnast/pull/6#issuecomment-163818469>.
>



-- 
Sarah Gripshover, PhD
Program Manager
PERTS
Department of Psychology
Stanford University

PR #7, opened 2015-12-11
fixing util.recode

daveponet:
@sarahgripshover here’s my suggested fix to util.recode


sarahgripshover:
This new version great to me! I see your point about not just accepting that behavior as our lot in life and dealing with it. I did find one unrelated problem which was that the value replacement section appeared twice. I deleted the duplicate section & will merge now.

PR #8, opened 2015-12-15
Package

cmacrander:
@daveponet @dkgreene @sarahgripshover gymnast is now a package!

See the very short and readable installation section of the readme. Basic idea: rather than `source()` from github, use Wickham's `devtools` to install the package from github.

Advantages:

* Namespacing: we don't have to call everything `util.*`, it acts like any other package
* Dependencies: normal installation/requiring takes care of this for us, no need for a fancy `gymnast_install()`
* File organization: don't need a monolithic `util.R` because of limited behavior of `source()`.
* Testing paradigm (not yet implemented)

**Note**: the installation procedure in the readme won't work as written until this is merged into master. Pre-merge, you have to target the `package` branch when installing from github:

    > devtools::install_github("PERTS/gymnast@package")

There's a lot more to clean up and re-structure, but the basic package workflow should already work (per above caveat).

Please critique!


daveponet:
@cmacrander this looks like a dream! I approve the merge.

cmacrander:
Dan says he's super buys, and Dave says he approves, so I call that reviewed.

PR #10, opened 2016-02-09
Read csv files

sarahgripshover:
Chris, this pull request consists of 2 commits:

* I deleted all the duplicated functions in this script (doesn't really belong in this branch but also pretty inoffensive so I didn't bother going back to Master to do it)
* I pasted the util.read_csv_files from the gist into gymnast, following your suggestions.

I'm now testing it to make sure it works, but I didn't change anything major so it should.


R/util.R
```diff
@@ -422,7 +422,50 @@ util.reverse_likert <- function(v, scale_levels) {
...
+          "The following files were not found: " %+%
+          paste0(
+              names(found_files)[!found_files],
+              collapse=", "
+              )
+        )
+        } else{
+            util.passed("All files present!")
+        }
+    # Note: I transform the strings "NA", "NULL", "", and " " into NAs when I import.
```
cmacrander: This comment no longer applies, correct?

R/util.R
```diff
@@ -422,7 +422,50 @@ util.reverse_likert <- function(v, scale_levels) {
...
+    for(file_name in names(path_list)){
+      # only try to read in files that exist!
+      if(file_name %in% names(found_files[found_files])){
+          read.csv(
+              path_list[[file_name]],
+              stringsAsFactors=FALSE,
+              ...
+              ) %>%
+              assign(file_name, ., environment)
```
cmacrander: This indentation pattern is confusing. Close brackets of all types should be vertically aligned with their invocation, so more like this:

```
read.csv(
    path_list[[file_name]],
    stringsAsFactors=FALSE,
    ...
) %>%
assign(file_name, ., environment)
```

That way I can quickly see that `read.csv()` and `assign()` are siblings in the same pipe.

R/util.R
sarahgripshover: lol no, just deleted it

cmacrander:
cool, merge it.

R/util.R
sarahgripshover: fixed!

PR #11, opened 2016-02-10
added hash_variable

sarahgripshover:
I simply added the hash_variable function. 2 things about it:

* I tried with `sapply` instead of `lapply` with `unlist` and it gave a named vector as the output, rather than a nice clean regular vector. I don't know whether the named vectors will actually mess anything up but figured `lapply` isn't that bad anyway.
* I started a new section in util.R because I imagine we'll have more of these hashing/de-identification related functions in the future.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -424,6 +424,17 @@ util.reverse_likert <- function(v, scale_levels) {
 
 ###############################################################
 ###
+###     De-identification
+###     Hash and de-identify variables and data.frames
+###
+###############################################################
+
+hash_variable <- function(x){
+    lapply(x, function(x) digest(x, algo="sha256")) %>% unlist
+}
+
+###############################################################
+###
 ###     Reading and writing files
 ###     Read/write files efficiently
 ###
--- a/R/util.R
+++ b/R/util.R
@@ -429,7 +429,9 @@ util.reverse_likert <- function(v, scale_levels) {
 ###
 ###############################################################
 
-hash_variable <- function(x){
+util.hash_vector <- function(x){
+    # Hashes vector x with SHA-256. Any salting must be done previously, this
+    # function only hashes.
     lapply(x, function(x) digest(x, algo="sha256")) %>% unlist
 }
 


cmacrander:
All the other functions are named `util.X`. I suggest `util.hash` for this one, or maybe `util.hash_vector`.

Also, add a little docstring to the first line of the function, e.g.

```
util.hash <- function (x) {
  # Hashes vector x with SHA-256.
  ...
}
```

sarahgripshover:
oh duh -- will change the function name to `util.hash_vector` and add docstring...and then, good to merge?

daveponet:
Do we ever want to hash *without* salting? If we commonly (exclusively?) do both at the same time, it might be nice for them to be handled by a single function -- both to save space and to standardize how the salt gets appended to the values, i.e., "starting at the end of the values, without any space."

sarahgripshover:
Seems to me it would be useful to have the option to do hashing without salting, because, well... you never know. Maybe there could be an optional salt parameter and if it's defined, then the variable is salted too.

I'm also working on a function called hash_data_frame that salts, hashes, removes the original identifier and optionally sorts by hashed id (to break row order correspondences).

```
hash_data_frame <- function(df, id_variable, salt, new_variable_name = NULL, sort_by_hashed = TRUE){
    # df is the data.frame to be de-identified
    # id_variable is the variable to be salted and hashed
    # salt is the salt
    
    # salt the id variable
    df[[id_variable]] <- paste0(df[[id_variable]], salt)
    
    # hash the salted variable
    df[[id_variable]] <- util.hash_vector(df[[id_variable]])
    
    # rename the id variable
    # if there's not a user-specified new variable name, paste the suffix "_hashed" onto the old one
    if(is.null(new_variable_name)){
        new_variable_name <- paste0(id_variable, "_hashed")
    }
    names(df)[names(df) %in% id_variable] <- new_variable_name
    
    if(sort_by_hashed){
        df <- df[order(df[[new_variable_name]]), ]
    }
    return(df)
}
``` 

Could put that in gymnast too if ppl think it's useful.

daveponet:
I love code reviews. Responses below. 

> Seems to me it would be useful to have the option to do hashing without salting, because, well... you never know. Maybe there could be an optional salt parameter and if it's defined, then the variable is salted too.

Yeah, I agree. I think an optional salt parameter would suit nicely.

Re: `hash_data_frame`, to me it seems too specialized for gymnast, at least for now.

cmacrander:
I concur with all and have no further comments. :rocket:

PR #12, opened 2016-02-10
modified util.apply_columns and added util.mutate_many

sarahgripshover:
As we discussed today, I modified `util.apply_columns`. However, I did it a bit differently than how we discussed:
- I did add a parameter to `util.apply_columns` allowing the user to restrict the function to select columns, because I think a person using this function a la carte might appreciate that parameter.
- However, I made `util.apply_columns` always, only return the columns to which the function was applied.
- I then wrote a separate function (which I've called `util.mutate_many` but I'm not married to that) to handle the logic of all the various ways you might want to use `util.apply_columns`. For example, you might want to replace the old columns in the data.frame with the modified ones or not, or you might have new column names you want to use or not.

My reasoning in breaking it up this way was that util.apply_columns seems like a nice, simple function with predictable output, whereas the logic of replacing/renaming columns is more complicated. But I could also just combine these two functions -- I'm interested to hear what you think.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -59,7 +59,7 @@ util.apply_columns <- function(df, fun, col_names = NULL, ...){
 util.mutate_many <- function(df,
     col_names,
     fun,
-    append = TRUE,
+    replace = FALSE,
     newcol_names = NULL,
     ...){
     # imitates the behavior of mutate in that it returns a data.frame of the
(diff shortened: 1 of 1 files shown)


cmacrander:
:sparkling_heart:

I think this is a great case study for writing unit tests. I'm thinking workshop.

sarahgripshover:
Yes! Let's do a workshop on unit tests! Would love to use this as a case study.

daveponet:
@sarahgripshover and @cmacrander, I have some reservations about this pull request. 

## Preface
I thought we are all on the same page that gymnast functions should be simple, clean, robust, and predictable. I believe this pull request takes us away from that ideal: It bloats function arguments in order to accommodate special-cases that could (and I very much believe should) be handled in other ways. 

## Objection 1
Re: "allowing the user to restrict the function to select columns". This could be handled with R's existing infrastructure, i.e., column indexing. That means a simpler, cleaner function with more predictable output. 

```r
cols_to_change <- c("col1","col2")
mutated_df[ , cols_to_change ] <- util.apply_columns( mutated_df[ ,cols_to_change ] )

```
Please do not integrate this change into gymnast. 


## Objection 2  (or Perhaps Just a Request for Clarification)
I'm confused about what `mutate_many` is supposed to do. What things that it's doing cannot be handled by using one or both of the following?

1. Call util.apply_columns on any subset of columns using column indexing (as above).
2. dplyr can be used with a vector of strings to mutate them selectively via the [dots parameter](http://www.r-bloggers.com/dplyr-use-cases-non-interactive-mode/). 

I'm guessing that there is some scenario that mutate_many handles that isn't captured in the above, but I'd love to review what that is and whether there's a more elegant way to handle it. I'm pretty uncomfortable with the copious arguments and hard-to-predict renamings of variables.

cmacrander:
> I thought we are all on the same page that gymnast functions should be simple, clean, robust, and predictable.

We _are_ on the same page about that. This implies we are somehow messing this up on purpose. We're not, and it's important to keep such negativity out of the code review process.

What you're really saying is, hey guys, 1) this isn't simple and predictable enough and 2) I know how we can be _even more_ simple and predictable. Both of these are very valuable comments and exactly why we do code review!

daveponet:
Thanks for the tone check, @cmacrander. I'm glad we're all on the same page about what gymnast is about.

PR #13, opened 2016-02-15
Read csv files

sarahgripshover:
# Background & implementation

`read_csv_files` is a function to read files from a list of path names. Dave pointed out that reading into global name space made him nervous, so we agreed to have `read_csv_files()` return a list of data.frames, and then have another function (`util.assign_list_to_environment`) do the assignment piece, resting a bit easier knowing that the name of the function suggests that the function will modify the workspace outside the function.

# Testing

I used the new functions in my current analysis script and also with some very simple test data and they worked as expected.

@daveponet I assigned Chris as the reviewer but I'd like you to sign off as well before we merge.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -455,15 +455,16 @@ util.read_csv_files <- function(path_list, environment = .GlobalEnv, ...){
     } else{
         util.passed("All files present!")
     }
+    # Read the new files into df_list()
+    df_list <- list()
     for(file_name in names(path_list)){
       # only try to read in files that exist!
         if(file_name %in% names(found_files[found_files])){
-            read.csv(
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -430,40 +430,56 @@ util.reverse_likert <- function(v, scale_levels) {
...
+            paste0(
+                names(found_files)[!found_files],
+                collapse=", "
+            )
         )
-        } else{
-            util.passed("All files present!")
-        }
+    } else{
+        util.passed("All files present!")
```
cmacrander: How about a touch more information here, e.g. `"All files present, successfully loaded " %+% length(found_files)`

cmacrander:
Looks good!

PR #14, opened 2016-02-18
added util.duplicated

sarahgripshover:
# Background

`duplicated()` in R's base is highly counterintuitive and not all that useful -- it returns TRUE for the _first_ time it sees a value, and then FALSE for all subsequent times it sees the value. e.g., `duplicated(c(1,1))` returns `FALSE TRUE`. What would be more useful and more intuitive is if there was a function that returns TRUE/FALSE based on whether the value is duplicated anywhere in the object; i.e., that returns `TRUE TRUE` for `c(1,1)`.

# Implementation

@dkgreene wrote a simple function a long time ago called `duplicated_or_original`. I copied it as-is and changed the name to `util.duplicated`.

# Testing

Dan and I have used this function for almost a year and it's never failed us, but I didn't do any special testing for this pull request.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -43,9 +43,12 @@ util.apply_columns <- function(df, fun, ...){
     )
 }
 
-util.duplicated <- function(x) {
-  # returns a boolean of length x indicating whether each value in x is
-  # duplicated anywhere in x
+util.duplicated_all <- function(x) {
+  # Returns a boolean indicating whether each value is
+  # duplicated anywhere in x. This differs from base::duplicated, which
+  # returns FALSE for the first of a set of duplicated values.
+  #
+  # Example: util.duplicated(c(1, 2, 1)) returns c(T, F, T).i
   duplicated_downward <- duplicated(x)
   duplicated_upward <- duplicated(x, fromLast=TRUE)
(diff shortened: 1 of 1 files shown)


cmacrander:
My only objection is to the name. It's sufficiently different in behavior from `base::duplicated` that I think it should be called something else, although I don't have any brilliant ideas... `is_duplicated`, `nonunique`, something like that.

R/util.R
```diff
@@ -43,6 +43,14 @@ util.apply_columns <- function(df, fun, ...){
     )
 }
 
+util.duplicated <- function(x) {
+  # returns a boolean of length x indicating whether each value in x is
+  # duplicated anywhere in x
```
cmacrander: Nit: it's nice to include proper english grammar, punctuation, and capitalization wherever possible. They make prose more readable, they also make comments more readable.

Furthermore, the difference between this and `base::duplicated` is subtle enough to warrant comment. I suggest:

```
  # Returns a boolean indicating whether each value is
  # duplicated anywhere in x. This differs from base::duplicated, which
  # returns FALSE for the first of a set of duplicated values.
  #
  # Example: util.duplicated(c(1, 2, 1)) returns c(T, F, T).
```

PR #15, opened 2016-03-10
Clean qualtrics question types

sarahgripshover:
# Background

Amy from CTC pointed out that a "duplicate column names" warning was getting thrown when using `qc.qualtrics_cleaning` on qualtrics files with a matrix table data type. This was happening because Qualtrics creates duplicate column names for matrix table rows, and distinguishes each column with an underscore + number suffix (e.g., `Q1_1`).
# Implementation

I replaced a specific regex search for the "_TEXT" suffix with a regex that finds any suffix.
# Testing

All the demo files were how I tested to convince myself this was working. It's just a visual comparison of original column names, pdd tag contents, and new column names, and an explicit test that column names are unique, despite the fact that pdd tags are not unique. I do think there should be a demo file but this is pretty ad-hoc and incomplete. I'm totally happy to delete it once the pull request is reviewed. (I included it now so Chris can see how I tested it.)


cmacrander:
Check out this example. Is this expected behavior or not?

```
df <- data.frame(Q1=c(1,2), Q2_1=c(1,2), Q2_2=c(1,2), Q3_1_1=c(1,2), Q3_1_2=c(1,2))

column_names <- c("Q1", "Q2", "Q2", "Q3_1", "Q3_1")

new_columns_names <- qc.add_suffixes(df, column_names)
# [1] "Q1"       "Q2_1"     "Q2_2"     "Q3_1_1_1" "Q3_1_1_2"
```

rumenPERTS:
I really like having a separate demo file. This makes it way more easier to see what exactly is the goal of the change.

I run the script, and I played with the regex myself and I didn't find any problems.

PR #16, opened 2016-03-11
extremely minor change: fixed util.print_pre and util.html_table to work in for-loops

sarahgripshover:
I made an extremely minor change to util.html_table and util.print_pre. When in interactive mode only, instead of simply returning the data object, I have these functions print the object. The reason is because `return()` does not cause objects to get printed to the console in the context of for-loops, which is annoying. So I changed `return()` to `print()` for interactive mode only -- that's literally all I did.

Rumen and I couldn't imagine a case when you'd want to use either of these functions for variable assignment that would truly require a return, but I want to run it by you in case you had a reason for returning the object that I'm not thinking of.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -169,11 +169,14 @@ util.as_numeric_if_number <- function(x){
 
 util.print_pre <- function(x){
     # prints to html as it would to console (preformatted html)
-    if(interactive()) return(x)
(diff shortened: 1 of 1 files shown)


daveponet:
:+1:

PR #17, opened 2016-03-31
Check index is unique

sarahgripshover:
# Overview

Function to quickly check whether a set of columns (`index_cols`) define unique rows.
# Implementation

It's pretty simple, just looks at the number of rows in the data.frame defined by unique combinations of the `index_cols` and compares that to the number of rows in the entire data.frame. If the index columns define unique rows, those two numbers should be the same!
# Testing

I have used this function extensively and it has never failed me, but I have not written systematic tests for it.

@daveponet , @cmacrander nominated this for gymnast, but I'd like you to sign off as well to make sure it meets your simplicity criteria!

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -378,7 +378,7 @@ util.rbind_union <- function(dfs){
 # util.rbind_union(dfs)
 # util.rbind_intersection(dfs)
 
-check_index_is_unique <- function(df, index_cols = NULL){
+util.check_index_is_unique <- function(df, index_cols = NULL){
     # Checks whether the vector of column names index_cols define unique rows
     # in df by comparing the number of rows in the whole df to the number of
     # rows defined by just the unique combinations of the index_cols. Returns
@@ -391,7 +391,7 @@ check_index_is_unique <- function(df, index_cols = NULL){
 
     # Throw an error if df is missing any index_cols
     if(any(!index_cols %in% names(df))){
-        stop("in check_index_is_unique, at least one of your index " %+%
+        stop("in util.check_index_is_unique, at least one of your index " %+%
             "cols does notexist in df.")
     }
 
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -378,6 +378,32 @@ util.rbind_union <- function(dfs){
...
+
+    # if no index cols are specified, then check whether the whole df is unique
+    if(is.null(index_cols)){
+        index_cols <- names(df)
+    }
+
+    # Throw an error if df is missing any index_cols
+    if(any(!index_cols %in% names(df))){
+        stop("in util.check_index_is_unique, at least one of your index " %+%
+            "cols does notexist in df.")
```
daveponet: Nit: Typo in "notexist"

R/util.R
```diff
@@ -378,6 +378,32 @@ util.rbind_union <- function(dfs){
...
+    }
+
+    nrows_index_cols <- df %>%
+        select(one_of(index_cols)) %>%
+        unique %>%
+        nrow
+
+    nrows_whole_df <- nrow(df)
+
```
daveponet: Nit: I think a more intuitive way to do this would be...

```r
any( duplicated( df[,index_cols] ) )

```

R/util.R
```diff
@@ -378,6 +378,32 @@ util.rbind_union <- function(dfs){
 # util.rbind_union(dfs)
 # util.rbind_intersection(dfs)
 
+util.check_index_is_unique <- function(df, index_cols = NULL){
```
daveponet: I think the name of the function could be a little clearer. When you say TRUE in response to check_index_is_unique, that presumably means the index is unique. But couldn't it also mean that you _checked_ if the index is unique? I might rename to `util.index_is_unique` to remove this admittedly small amount of ambiguity.

daveponet:
Cool. I agree it should go in gymnast. :+1:

PR #18, opened 2016-04-12
listing files and finding in crypts

cmacrander:
@daveponet fyi: tool for making it easier to find you data files in mounted crypts, regardless of which slot you have them in, for immediate use with CTC analysts.

## Background

People are used to operating on deidentified data in the R scripts. These are not in crypt files, so the path to the data is static, e.g. `~/Dropbox/PERTS shared data/something/something`. We're changing things up and making everyone keep everything in crypts. This is annoying because the path to your data can change depending on which slot you mount it in.

## Implementation

This code solves that problem. Assuming 1) you're not using a Windows computer 2) your VeraCrypt mounts things in `/Volumes/NO NAME[ X]`, you just need to give the file name in order to load the file.

The function is designed to operate smoothly with util.read_csv_files, so you can write something like this:

```
my_data <- util.read_csv_files(util.find_crypt_files(list(
  cc10 = "CC10/all_students.csv",
  ku51 = "KU/5-1/cool_stuff.csv"
)))
```

You only need to write as much of the file path as is necessary to uniquely identify the file, and the files might be spread across arbitrary mount points.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -608,12 +608,9 @@ util.find_crypt_paths <- function (files_to_load, initial_path = NA,
     for (label in names(files_to_load)) {
         file_name <- files_to_load[[label]]
 
-        # Make sure the file name starts with a slash so that, when grepping, we
-        # don't match in the middle of a file or folder name.
-        if (substr(file_name, 1, 1) != '/') {
-            file_name <- '/' %+% file_name
-        }
-        match <- crypt_files[grepl(file_name, crypt_files)]
+        # fixed = TRUE means interpret the file name as a literal subset to match,
+        # not a regular expression (where characters like . have special meaning).
+        match <- crypt_files[grepl(file_name, crypt_files, fixed = TRUE)]
         if (length(match) > 1) {
             stop("Multiple matches found for " %+% file_name %+% ": " %+%
                  match %+% "\n")
--- a/R/util.R
+++ b/R/util.R
@@ -531,25 +531,17 @@ util.list_files <- function (initial_path, max_depth = 2, current_depth = 0) {
     # current_depth: internal use only, do not specify.
     #
     # Returns: char of absolute file paths
-
-    if (.Platform$OS.type != 'unix') {
-        stop("util.list_files() only supports unix-like systems, not windows.")
-    }
-
+    
+    
     # List everything within this path, both files and dirs.
-    all_names <- list.files(initial_path, full.names = TRUE)
-
-    # List directories within this path.
-    dirs <- list.dirs(initial_path, recursive = FALSE)
-
-    # Exclude any hidden or system files.
-    ignore_pattern <- '/[\\.\\$~]'
-    all_names <- all_names[!grepl(ignore_pattern, all_names)]
-    dirs <- dirs[!grepl(ignore_pattern, dirs)]
-
-    # Figure out what the real files are by excluding dirs.
-    files <- setdiff(all_names, dirs)
-
+    all_names <- list.files(initial_path, pattern = '^[^\\.\\$~]',
+                            full.names = TRUE)
+    
+    # file.info() returns a data frame, use it to separate files and dirs.
+    info <- file.info(all_names)
+    files <- all_names[!info$isdir]
+    dirs <- all_names[info$isdir]
+    
     # If not at max depth, recurse into each found directory.
     if (current_depth < max_depth) {
         for (d in dirs) {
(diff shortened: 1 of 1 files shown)


rumenPERTS:
One thing I am not sure about is if you assume that mount paths will always contain /NO NAME
With True Crypt I've seen paths like "NO_NAME" too.

cmacrander:
I tried to investigate the "NO NAME" pattern and didn't find much. It appears to be a feature of OS X, not TrueCrypt or VeraCrypt. When a volume is mounted without an explicit name, that's what happens. When mounting, it's possible to specify a specific directory, but we don't have any reason to do that.

I suspect this will only be a problem if someone is using linux. Googling again, it seems they might be on paths like `/media/veracrypt1`.

Maybe assuming OS X is bad for gymnast. I'm eager for suggestions on how to search for file paths in a more general way.

R/util.R
```diff
@@ -514,6 +514,107 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+    # * Lists files, not directories. Because it's called list_files.
+    # * Rather than choosing between no recursion (not very useful) and full
+    #   recursion (potentially very slow for a deep folder tree) you can set a
+    #   max depth. Choose 0 for no recursion.
+    # * Ignores hidden and system files, defined as anything starting with
+    #   '.', '$', or '~'.
+    # * Only works on unix-like systems (no windows)!
+    #
+    # Args:
+    #   initial_path: atomic char, directory to scan for files.
```
sarahgripshover: will this always be "/Volumes/"?

R/util.R
```diff
@@ -514,6 +514,107 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+    dirs <- dirs[!grepl(ignore_pattern, dirs)]
+
+    # Figure out what the real files are by excluding dirs.
+    files <- setdiff(all_names, dirs)
+
+    # If not at max depth, recurse into each found directory.
+    if (current_depth < max_depth) {
+        for (d in dirs) {
+            files <- c(files, list_files(
+                d, max_depth = max_depth, current_depth = current_depth + 1))
```
sarahgripshover: Why is this `list_files` instead of `util.list_files`? I don't see `list_files` defined anywhere...is this some magic that I don't know about yet?

R/util.R
```diff
@@ -514,6 +514,107 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+    #   initial_path: atomic char, default '/Volumes', the parent directory
+    #     where crypt files are mounted.
+    #   max_depth: atomic int, default 2, how many subfolders deep to scan for
+    #     files. Zero means enter no subfolders.
+    #
+    # Returns: List with provided labels to absolute file paths.
+
+    all_volume_paths <- list.dirs(initial_path, recursive = FALSE)
+    mount_paths <- all_volume_paths[grepl('/NO NAME', all_volume_paths)]
+
```
sarahgripshover: sometimes the mounted files are called 'untitled' instead of 'NO NAME'

R/util.R
cmacrander: Nope, that's just a bug. Good catch.

R/util.R
cmacrander: Really? Then there's something I don't understand. I couldn't make my computer do that. Maybe you can show me when you have a chance.

R/util.R
```diff
@@ -514,6 +514,107 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+    # Compile a list of files from each mount path.
+    crypt_files <- c()
+    for (m in mount_paths) {
+        crypt_files <- c(list_files(m), crypt_files)
```
cmacrander: same renamed-function bug here

daveponet:
Generally happy with this direction! Note this needs to work on Windows computers too.

R/util.R
```diff
@@ -514,6 +514,107 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+    }
+
+    # For each file to load, scan the list of known files for a match.
+    found_paths <- list()
+    for (label in names(files_to_load)) {
+        file_name <- files_to_load[[label]]
+
+        # Make sure the file name starts with a slash so that, when grepping, we
+        # don't match in the middle of a file or folder name.
```
sarahgripshover: that's just in case the user gives files that don't start with / ?  Might be helpful to note in the comments section that the files_to_load has to contain full file and/or directory names, not partial ones, e.g., in case someone like me gets the bright idea to try to use "college/data.csv" because I don't know whether "data.csv" is in the "community_college" directory or the "all_college" directory.

sarahgripshover:
ok, looks good to me! Just made a few minor comments. The most important is just to note that some volumes can be labeled "untitled" not "NO NAME". I'd say go ahead and merge, is there a plan for testing?

cmacrander:
@sarahgripshover I addressed some of your points. Now you can be clever and start your file path match right in the middle of some folder name if you want to. I added the "Untitled" volume possibility, and made the volume names configurable if someone has weird ones.

R/util.R
sarahgripshover: yeah, I dk anything about why that happens. Some of my crypts default to "NO NAME" and others default to "untitled". I think the KU ones default to untitled, for example. I have no idea why one thing happens rather than the other.

PR #19, opened 2016-04-22
changed na.rm = TRUE

sarahgripshover:
This is another extremely minor tweak but want @daveponet to review in case it interferes with your vision for these functions. Basically, I want to use util.row_means but can't because I don't want to set na.rm = TRUE. So I changed both util.row_means and util.row_sums so that you can set na.rm in the function call. Still has default TRUE value so won't break other things. @cmacrander probably good for you to eyeball this too!


cmacrander:
Looks good to me!

PR #20, opened 2016-04-22
List files

cmacrander:
fyi @daveponet 

As I started to use these listing functions, I realized I needed more flexibility to find files sometimes, directories sometimes, and sometimes both. This cleans up the functions so they can be used more flexibly.

`util.list_all()` lists both files and directories within `max_depth` of some initial path.
`util.list_files()` is the same, but lists just files.
`util.list_dirs()` is the same, but lists just directories.

`util.find_crypt_paths()` now can also find directories within crypts, not just files.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -514,42 +514,65 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
     }
 }
 
-util.list_files <- function (initial_path, max_depth = 2, current_depth = 0) {
+util.list_all <- function (initial_path, max_depth = 2, type = 'all', current_depth = 0) {
     # Like base::list.files, but better:
-    # * Lists files, not directories. Because it's called list_files.
+    # * Can choose to list only files, not directories (type = 'files').
     # * Rather than choosing between no recursion (not very useful) and full
     #   recursion (potentially very slow for a deep folder tree) you can set a
     #   max depth. Choose 0 for no recursion.
     # * Ignores hidden and system files, defined as anything starting with
     #   '.', '$', or '~'.
-    # * Only works on unix-like systems (no windows)!
     #
     # Args:
     #   initial_path: atomic char, directory to scan for files.
     #   max_depth: atomic int, default 2, how many subfolders deep to scan for
     #     files. Zero means enter no subfolders.
-    # current_depth: internal use only, do not specify.
+    #   type: atomic char, default 'all', or choose 'dirs' or 'files'.
+    #   current_depth: internal use only, do not specify.
     #
     # Returns: char of absolute file paths
     
+    # Remove the trailing slash if it exists.
+    len <- nchar(initial_path)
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -514,42 +514,65 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
     #   '.', '$', or '~'.
-    # * Only works on unix-like systems (no windows)!
     #
     # Args:
     #   initial_path: atomic char, directory to scan for files.
     #   max_depth: atomic int, default 2, how many subfolders deep to scan for
     #     files. Zero means enter no subfolders.
-    # current_depth: internal use only, do not specify.
+    #   type: atomic char, default 'all', or choose 'dirs' or 'files'.
+    #   current_depth: internal use only, do not specify.
```
sarahgripshover: Out of curiosity, what would happen if a user tried to specify current_depth? Seems like it would cause things to break. What is the reason for including it as a function argument that the user is told not to specify, rather than just setting current_depth <- 0 at the beginning of the function or maybe around line 557?

sarahgripshover:
Looks good to me! Just made one comment, and another Q: is there any reason at this point to avoid generalizing util.find_crypt_paths beyond just finding crypt paths, to finding paths in any directory? Obviously finding crypt paths the main use case at this point, but at least for the Mac OS/Unix version it seems like you could generalize the find_crypt_paths function to any directory, not just "/Volumes", pretty easily. Probably not worth doing if it involves much effort but I could imagine it being more useful to our hypothetical gymnast users if it worked in any directory.

R/util.R
cmacrander: That's how I designed the recursion. Each time it recurses, it passes the current depth to the next call; that way it knows when to stop. You're not supposed to set it b/c this function has two modes: a human calling it, and it calling itself. The parameter is only used in the second case.

cmacrander:
You could just grep over the results of the list functions to accomplish that, which is was `find_crypt_paths` does. Finding crypt paths deserves its own special function because of the special rules about where we can expect that crypt files will be mounted.

R/util.R
```diff
@@ -514,42 +514,65 @@ util.assign_list_to_environment <- function(l, environment = .GlobalEnv){
...
+
+util.list_files <- function (initial_path, ...) {
+    util.list_all(initial_path, type = 'files', ...)
+}
+
+util.list_dirs <- function (initial_path, ...) {
+    util.list_all(initial_path, type = 'dirs', ...)
 }
 
 util.find_crypt_paths <- function (files_to_load, initial_path = NA,
```
cmacrander: needs `max_depth` in the signature

R/util.R
sarahgripshover: ahh that makes sense!

sarahgripshover:
ok that makes sense too!

PR #21, opened 2016-08-02
repairng _TEXT

rumenPERTS:
There was a bug which repeats the string _TEXT if we use the original
column name (first row in qualtrics). In the current fix we first check
if the column name is old or new, and adds _TEXT only if it is new
column name. This is pretty much verbatim of the solution that Dave
proposed in github. Tested, and it works fine.


sarahgripshover:
Looks good to me!

PR #22, opened 2016-08-04
initial commit

sarahgripshover:
Commit includes scale computation function (util_scale_computation.R) and test document (demos/scale_comp_informal_demo.R). For now
it’s just an informal demo but can be made more formal when module is
ready for prime time & other functions are computed.

For now, I’ve only written the sc.append_scales function because I need
that right now. I'll work on the summary functions next.

As for feedback on this round, this is going in gymnast so definitely worth it to nitpick the code! @daveponet would love to hear from you too if you have a second!


R/util_scale_computation.R
```diff
@@ -0,0 +1,90 @@
...
+        ]
+
+        # look up the corresponding column names in survey_df
+        survey_scale_column_names <- names(survey_df)[
+            names(survey_df) %in% scale_items
+        ]
+
+        missing_items <- scale_items[
+                !scale_items %in% survey_scale_column_names
+            ]
```
cmacrander: nit: indents are off

R/util_scale_computation.R
```diff
@@ -0,0 +1,90 @@
...
+        if(length(survey_scale_column_names) > 0){
+            # save the existing column names to use for renaming variables
+            # later
+            old_names <- names(survey_df)
+            # compute and append row means
+            scale_means <- util.row_means(survey_df[survey_scale_column_names])
+            survey_df <- data.frame(
+                survey_df,
+                new_scale = scale_means
+            )
```
cmacrander: why not just `survey_df[[new_scale]] <- util.row_means(survey_df[survey_scale_column_names])`

R/util_scale_computation.R
```diff
@@ -0,0 +1,90 @@
...
+            if(add_z_score){
+                scale_means_z <- util.z_score(scale_means)
+                survey_df <- data.frame(
+                    survey_df,
+                    new_scale_z = scale_means_z
+                )
+                # rename the zscore scale variable with the suffix "_z"
+                names(survey_df)[
+                    names(survey_df) %in% "new_scale_z"
+                    ] <- scale %+% "_z"
```
cmacrander: again, `survey_df[[scale %+% "_z"]] <- util.z_score(scale_means)`

cmacrander:
I don't know what the right call is here, but I'm suspicious of the use of `util.warn()` here. Would it be more appropriate to actually throw an error, i.e. use `stop()` instead? It depends on how serious a problem these things are.

R/util_scale_computation.R
sarahgripshover: lol because I always forget you can do that!

R/util_scale_computation.R
sarahgripshover: yep

R/util_scale_computation.R
sarahgripshover: sorry!

sarahgripshover:
oh sorry, I missed your util.warn column. In this case, it's not a serious problem if items are missing and even if no scale is computed. In fact, I imagine it will frequently be the case that we're using a variables table with variables that are missing in a particular survey dataset. I would not want the scale computation function to break in this case. Does that make sense?

I'm going to go ahead and merge this, but lmk if you want to discuss it further!

-SG

PR #24, opened 2016-09-06
build and summarize glm models

daveponet:
I'm only officially requesting review from @cmacrander and @sarahgripshover. However, I encourage review and suggestions from all @PERTS/analysts.

## Goals
The goal of the `ds` module is broadly to:

* Make it fast and easy to build models and summarize data.
* Standardize the data summaries we create for ourselves and collaborators. 

This pull request is focused on building and summarizing glm models. It stops short of creating standard, readable data tables from those models. However, it introduces upstream functions in the hope is that it will be much easier for us to build presentation-ready tables once we have functions that make it fast and easy to generate models and extract summary information from them.

## Notes to Reviewers
I'm sure there's a lot to improve about this code, and I'm looking forward to feedback.
@PERTS/analysts, I'm especially excited for how you think we should build summary tables and graphs from this output.


R/util_data_summaries.R
```diff
@@ -1,71 +1,63 @@
...
-    obs_max = function(x){
-        if( util.is_vector_of_numbers(x) ){
-            return(max(as.numeric(x),na.rm=T))
-        }else{
-            return(NA)   
-        }
+ds.helper <- list()
+
+ds.helper$mean <- function(x){
+    if( util.is_vector_of_numbers(x) ){
```
cmacrander: The Wickham says: http://adv-r.had.co.nz/Style.html
PERTS style guide: https://docs.google.com/document/d/1nO0rFe3MHPDE6gwpHhzFH5PGBV8D5gqtuNXbBgtwL2M/edit

This line, and others like it, should be:

`if (util.is_vector_of_numbers(x)) {`

R/util_data_summaries.R
```diff
@@ -1,71 +1,63 @@
...
+    mean(util.is_blank(x)) * 100
+}
+
+ds.helper$sd = function(x){
+    sd(x, na.rm=TRUE)
+}
+
+ds.helper$obs_min = function(x){
+    if( util.is_vector_of_numbers(x) ){
+        return(min(as.numeric(x),na.rm=T))
```
cmacrander: `T` vs. `TRUE` should be consistent.

R/util_data_summaries.R
```diff
@@ -1,71 +1,63 @@
...
+}
+
+ds.helper$obs_max = function(x){
+    if( util.is_vector_of_numbers(x) ){
+        return(max(as.numeric(x),na.rm=T))
+    }else{
+        return(NA)   
     }
-)
+}
```
cmacrander: You could abstract several of these like this, if you thought it was helpful:

```
handle_nas <- function (f) {
  return(function (x) {
    if (util.is_vector_of_numbers(x)) {
      return(f(as.numeric(x), na.rm = TRUE))
    } else {
      return(NA)
    }
  }
}

ds.helper$mean <- handle_nas(mean)
ds.helper$sd <- handle_nas(sd)
ds.helper$obs_min <- handle_nas(min)
ds.helper$obs_max <- handle_nas(max)
```

R/util_data_summaries.R
```diff
@@ -77,17 +69,54 @@ ds.helper$default_col_funcs = list(
...
     "n_unique" = ds.helper$n_unique
-)   
+)  
 
+ds.helper$variable_type <- function(x){
+    # return variable type as
+    # (Note that NAs are ignored)
+    #  boolean - x contains exactly c(0,1) or c(TRUE,FALSE)
+    #  numeric - x is numbers 
+    #  categorical - x is a string or factor with up to 20 levels
```
cmacrander: I don't see the number 20 appearing in the function...

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+    ivs = c(),           # independent variables each run independently
+    mods = c(""),        # moderators each run independently, default no mod
+    covs = c(),          # covariates included all together or none
+    cov_groups = list() # customize cov groups instead of default all or none
+    
+){
+    # Build vector of model formulas that meet the ds.glm1 spec
+    # for each unique combination of dv x iv x mod x cov_group.
+    # If covs supplied, cov_group gets one entry for no covs (unadjusted) 
+    # and one for all covs supplied in covs argument (adjusted).
```
cmacrander: The string manipulation here is pretty complicated. Worth providing an example, e.g. with dvs = X and ivs = Y, creates the formula `"I ~ don't get ~ formula strings"`

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+    if(msl$family %in% "binomial"){
+        msl$iv_apa <- "OR=" %+% round(exp(msl$iv_coef),3) %+%
+            ", z=" %+% round(msl$iv_stat,3) %+%
+            ifelse(msl$iv_p < .001,", p<.001",", p=" %+% round(msl$iv_p,3))
+            
+    } else if(msl$family %in% "gaussian"){
+        msl$iv_apa <- "b=" %+% round(msl$iv_coef,3)  %+%
+            ", t(" %+% msl$df %+% ")" %+% "=" %+% round(msl$iv_stat,3) %+%
+           ifelse(msl$iv_p < .001,", p<.001",", p=" %+% round(msl$iv_p,3))
+    }
```
cmacrander: Sounds like `msl_to_apa_string()` should be its own function.

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+
+ds.helper$unit_test <- function(){
+    
+    assert_expected_output <- function( func_name , input, expected_output ){
+        # warn if evaluated func_name(input) != expected_output
+        # errors messages are not compared, but errors can be expected
+        
+        # func_name passed as string rathern than func itself
+        # so the func_name can be included warnings
+        func <- eval(parse(text=func_name))
```
cmacrander: Better to use `get()`:

`func <- get(func_name)`

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+        }
+        
+        if( ! ( both_are_errors | outputs_identical ) ){
+            "assert_expected_output failed in " %+%
+                func_name %+% " when passed '" %+%
+                paste(input,collapse=", ") %+% "'" %+% 
+                ". \n" %+% error_message %>%
+                util.warn()
+        }
+    }
```
cmacrander: Clever. However, we should move to using `Runit`.

cmacrander:
I can tell you had a blast writing this!

And you included both a demo and unit tests! 🏆 🎱 🌟 🎸 

From perusing the demo Rmd, I gleaned that the usefulness of this comes in when you want to run a bunch of simple models all at once. But I'm still left a little unsure that some of those more involved functions are really necessary. Are there really not any tools for extracting the variables from a set of models? Must you use laborious string manipulation to accomplish those things? I'm not sure I'm asking the right questions... I guess I would just like to see a little more justification for the approach.

R/util_data_summaries.R
sarahgripshover: Could the 20 levels cutoff be a parameter that defaults to 20? Might not want to be wed to it.

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+    if( sum( ! predictors %in% varlist ) > 0){
+        varlist$covs <- predictors[ ! predictors %in% varlist ]
+    }
+    
+    return(varlist)
+}
+
+
+ds.glm1 <- function( formula, data, mod_family=NULL ){
+    # run the appropriate model, unless mod_family override is supplied
```
sarahgripshover: nit: "mod_family" sounds like it could mean "moderator_family"—I'd just call it "model_family" for clarity.

R/util_data_summaries.R
```diff
@@ -77,17 +69,54 @@ ds.helper$default_col_funcs = list(
...
+    } else if( util.is_vector_of_numbers(x) ){
+        return("numeric")
+    }
+    
+    n_unique <- x[!util.is_blank(x)] %>%
+        unique() %>%
+        length()
+    
+    if(n_unique < 2){
+        return("invariant")
```
sarahgripshover: Nit: here you're making it so the user will get an "unsupported model type" output for dvs that are either categorical or invariant. But the way you've written it, the "invariant" warning only works for non-numeric and non-boolean DVs. If a DV were e.g., c(1,1,1), you would be able run the model just fine, and you'd get all zero coefficients and hopefully you'd realize something was up and check your DV at that point. But, if you've been using this package a lot, it might be confusing that you didn't get the "invariant" warning that you would get for c("dog", "dog", "dog"). So I'd recommend either returning "invariant" here for all dvs with < 2 unique values, or else just have "categorical" be the only unsupported variable type.

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+            is_valid <- FALSE
+        }
+    }
+    return(is_valid)
+}
+
+
+ds.build_glm1_formulas <- function(
+    meta_formula = NULL, # @todo: enable specification through formula 
+    variable_vectors = NULL, # @todo: meta_formula can reference named vectors
```
sarahgripshover: since these are unsupported they should come last—it's annoying to have to type all the argument names

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+    if( length(covs) > 0 ){
+        # if covs is supplied, use it to define adjusted and unadjusted
+        # covariate groups with all and none of the covs, respectively
+        cov_groups <- list()
+        cov_groups[["Unadusted"]] <- ""
+        cov_groups[["Adjusted"]] <- covs
+    } else if( length(cov_groups) == 0){
+        cov_groups <- list( Unadjusted = "" )
+    }
+
```
sarahgripshover: It's nice to have the "adjusted" and "unadjusted" info here (there's a typo btw, "unadjusted" is missing a "j"), but it gets dropped down the line. By the time you run `ds.model_summary_df`, you get a data.frame with all the formulas you built here, which could be a lot if you have a lot of ivs and dvs. Sometimes you might not care about the unadjusted models and want to drop them, or you might want to be able to sort them to the top or something like that. So it could be useful to propogate this information forward so that the final summary df has an "adjusted" column.

R/util_data_summaries.R
```diff
@@ -126,3 +155,509 @@ ds.summarize_by_column <- function(
...
+    # use of column and row numbers seems dangerous;
+    # however order should be consistent when 
+    # ds.glm1 spec is used
+    # also, irregularities between different model types
+    # e.g., t vs z for guassian and binomial families
+    # and level is appended to row.names when TRUE/FALSE
+    msl$iv_coef <- coefs[2,1]
+    msl$iv_se <- coefs[2,2]
+    msl$iv_stat <- coefs[2,3]
+    msl$iv_p <- coefs[2,4]
```
sarahgripshover: Here's a place where the purrr package could offer more flexibility and make it easier to e.g., extract coefficient info for multiple levels of a categorical dv

R/util_data_summaries.R
cmacrander: They're still optional, even though they're first!

sarahgripshover:
Wow, what a great start!

Things I love about the model summary table:

- It has an APA format column, lovely!!! SO useful!!!!
- I like the prefixes, makes it easy to see what is what in the summary (though I might change "mod" to "moderator" because it sounds like maybe it could stand for "model")

I also think the first priority in expanding this should be enabling categorical IVs.

R/util_data_summaries.R
sarahgripshover: Well actually, you could just sort the table by is.na(covs) so maybe nevermind.

daveponet:
Hot damn! I love code review! So many great nits, you guys!

🎸 🎸 🎸

R/util_data_summaries.R
daveponet: I tried this, and it doesn't work. I think `get` only retrieves complete objects in the workspace, not on their children. See below.

This works:
```r
get("ds.helper")
```

But this:
```r
get("ds.helper$variable_type")
```
Returns
```
Error in get("ds.helper$variable_type") : 
  object 'ds.helper$variable_type' not found
```

Anyway, I'm leaving this as is.

R/util_data_summaries.R
daveponet: I'm sure there are good arguments to be made for using `Runit`. Here are two possible reasons _not_ to refactor for Runit:

1. It's yet another package that `gymanst` would depend on. Generally speaking, more dependencies = bad.
    1. Packages sometimes get stale and break in new versions. If a package is popular and well-maintained and/or it does something really complicated that would take a long time to code, then we should obviously use it. (Aside: [RUnit is downloaded 8,000/month but is "no longer actively developed"](https://cran.r-project.org/web/packages/RUnit/README.html)). Regardless of the specific case here, I just want to mention for posterity (and I'm guessing you probably agree with me already) that our "default" should _not_ be to use an external package to do something that we can easily do ourselves. Personally, I think there are few worse things in coding than running code that's been working great for years and having it break because somebody else messed up a dependency. Then, instead of getting to do the analysis work you expected to (and are probably are on a deadline for), you have to dig through the bowels of your code and refactor it because somebody else, far away, messed up. 
    2. More packages mean slower load times and higher risk for name-space conflicts.
2. This is already written and appears to work. Rewriting it to use a different approach would take more time investment.


If you think there are big advantages to using RUnit, you're probably right. Maybe you could lead a workshop on it, and we could all together hash out pros and cons. (Please just make sure I can come to it because that's a conversation I wouldn't want to miss.)

R/util_data_summaries.R
daveponet: I am scandalized by this oversight! Great catch.

R/util_data_summaries.R
daveponet: For now, I'm going to leave them first because I fantasize about ds.glm to eventually looking and feeling like glm (where the formula argument comes first). E.g., I think it'd be nice to call this like so, and have it parse the [] into arguments.

```r

ds.build_glm1_formulas([DV1, DV2] ~ [IV1, IV2] * MOD + [COV1, COV1 + COV2])


```

daveponet:
@PERTS/analysts,

Huge thanks for excellent recommendations! It was great to build this new module with you all.

I'm pulling it in now. That means you should default to using ds.glm1 whenever statistically appropriate. That should help us write better analysis code and identify ways to improve the ds module.

❤️ ,
~Dave

PR #25, opened 2016-10-07
Fix util is vector of numbers

sarahgripshover:
This is a fix to a bug I found in util.is_vector_of_numbers. Basically, the regex it was using works perfectly for every numeric type except for scientific notation (e.g., 1e05). I realized it was dumb to use a regex anyway, and instead now it checks for numeric values with as.numeric(). I also added some unit tests to avoid missing these kind of problems in the future!


cmacrander:
Brilliant!

Yes, regexes are half-evil. They tempt you to use them in all sorts of ways where you really shouldn't. They're kind of like fire.

sarahgripshover:
yeah I think I was going through an overly-enthusiastic-about-regex phase. e.g., see this cartoon https://xkcd.com/208/

daveponet:
@sarahgripshover, this change breaks the function: It no longer ignores blanks (as required). 

"Are all elements of x numbers (permitting blanks)?"

We should probably add a check for that to the unit tests too....

sarahgripshover:
@cmacrander I finally addressed Dave's accurate objection to my fix from two years ago, and added a check for correct handling of blanks to the unit test. With your approval I will go ahead and merge with master.

sarahgripshover:
@chris-perts bumping this. Ok to merge?

R/util.R
```diff
@@ -130,27 +130,62 @@ util.is_vector_of_numbers <- function(x){
...
+    x_char <- c("1", "a", "b", "1.1", "1e05")
+    x_number <- c("1", "1.1", "2", "-1.1")
+    x_scientific <- c("1e05")
+    x_with_blanks <- c(x_number, "", NA)
+    x_logical <- c(TRUE, FALSE)
+
+    if(util.is_vector_of_numbers(x_char) == TRUE){
+        stop("In util.is_vector_of_numbers, vectors containing " %+%
+                 "character values are being evaluted as numbers.")
+    }
```
chris-perts: nit: I try to eliminate as many negatives and double negatives in code as possible, because humans are slow at reading them. The fact that the expression in the `if` condition is the opposite of what you're looking for ("stop if this is WRONG thing is true") threw me the first time. It's more normal, and requires fewer mental gymnastics, to make a positive statement like "I assert this is true". This is why, in my opinion, gymnast should define:

```
assert <- stopifnot
```

Then also note the error message of a failed `stopifnot` is actually pretty useful b/c it tells you the name of the argument you passed, so with your well-named inputs you can drop all the custom stop messages. Put it together, and the code becomes much more readable:

```
util.is_vector_of_numbers_test <- function(){
    x_char <- c("1", "a", "b", "1.1", "1e05")
    x_number <- c("1", "1.1", "2", "-1.1")
    x_scientific <- c("1e05")
    x_with_blanks <- c(x_number, "", NA)
    x_logical <- c(TRUE, FALSE)

    assert(!util.is_vector_of_numbers(x_char))
    assert(util.is_vector_of_numbers(x_number))
    assert(util.is_vector_of_numbers(x_scientific))
    assert(util.is_vector_of_numbers(x_with_blanks))
    assert(!util.is_vector_of_numbers(x_logical))
 }
 
 
util.is_vector_of_numbers_test()

# Assuming your fix wasn't in place:
# Error: util.is_vector_of_numbers(x_with_blanks) is not TRUE
```

Which I would read as something like "make sure the function correctly fails in this case" rather than "if this function fails to fail then stop."

R/util.R
```diff
@@ -130,27 +130,62 @@ util.is_vector_of_numbers <- function(x){
...
         return(TRUE)
     }
-    else{
-        return(FALSE)
+}
+
+
+util.is_vector_of_numbers_test <- function(){
+    x_char <- c("1", "a", "b", "1.1", "1e05")
+    x_number <- c("1", "1.1", "2", "-1.1")
```
chris-perts: nit: I can think of some other number-like strings we might want to check: `".1"`, `"0.1"`, `"00.1"`,  `"1.00"`, `"+1.1"`, `"01"`.

R/util.R
```diff
@@ -130,27 +130,62 @@ util.is_vector_of_numbers <- function(x){
...
+    
     # find numeric values
-    numeric_values <- grepl("^-*[[:digit:]]*\\.*[[:digit:]]+$",x)
+    x_as_numeric <- suppressWarnings(as.numeric(x))
 
-    # find the blank values
-    blank_values <- util.is_blank(x)
+    # anything that gets coerced to NA by as.numeric is not really a number.
+    # but elements that were blank originally are ok.
+    blanks <- is.na(x_as_numeric)
```
chris-perts: nit: based on your comments, I'm finding the use of `blank` in the variable names confusing. What do you think of this (logic should be identical):

```
non_numeric <- is.na(x_as_numeric)
originally_blank <- util.is_blank(x)
if(any(non_numeric & !originally_blank)) {
  ...
}
```

PR #26, opened 2016-10-13
Prevent name conflicts when used as package

cmacrander:
Conflict was with our` %+%` versus `` psych::`%+%` ``

This is relevant for codeship continuous integration, which uses
gymnast as a package.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -685,6 +685,21 @@ gymnast_install <- function () {
             library(lib_name, character.only = TRUE)
         }
     }
+    resolve_name_conflicts()
 }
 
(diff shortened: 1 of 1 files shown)


cmacrander:
When I patch this code into my environment, then your perts_analyses unit tests pass, Sarah. Will have no new effect on sourcing gymnast from github, since `%+%` is already in the environment, as we all know from using it all the time.

sarahgripshover:
This looks great, thanks for getting to the bottom of it! Also nice to have a framework for handling any future name conflicts.

PR #28, opened 2017-11-03
minor tweaks

sarahgripshover:
**Background & implementation**
Hey Chris, I made some extremely minor changes to gymnast. I almost considered not doing a pull request, but because so many people use gymnast, I figured I ought to have another pair of eyes on it in case I'm being a complete moron about something. The changes were to fix a couple of annoying things that were annoying me:
- in qc.insert_hidden_column_names, I added a suffix convention to differentiate any duplicate column names that were created by applying pdd tags. A warning is still thrown when this happens, but the suffixes render the resulting dataset usable, so that one can troubleshoot the duplicates if desired (or ignore if not desired)
- in qc.rbind_in_progress, I added a line of code to append an extra column to complete responses before rbinding them to in-progress responses, because Qualtrics includes an extra "in progress" column that breaks rbinding if you forget to add it to the complete responses.
- changed [,dv_variable] to [[dv_variable]] in ds.glm1 because in some version of R, the [,dv_variable] notation returns a matrix instead of a vector which messes everything up
- changed is.na to util.is_blank in sc.append_scales, just in case of non-NA blanks

**Testing**
These were such minor changes I didn't feel like they needed formal testing. I can say I've been working with them for a while and haven't had any problems. Ideally, we would already have unit tests running in the background, but I don't think we do and I haven't been able to find the time to write some.


R/util_data_summaries.R
```diff
@@ -373,7 +373,7 @@ ds.glm1 <- function( formula, data, family=NULL) {
     # run the appropriate model, unless family override is supplied
     dv_variable <- ds.helper$glm1_formula_to_varlist(formula)[["dv"]]
     if (is.null(family)) {
-        dv_var_type <- ds.helper$variable_type(data[,dv_variable])
+        dv_var_type <- ds.helper$variable_type(data[[dv_variable]])
```
chris-perts: This seems like it was definitely a bug before. Doesn't `[,dv_variable]` mean "subset this dataframe, taking all rows and the column named X" giving you a one-column df? If the goal is to extract the column as a vector, your new code is the way to go.

R/util_qualtrics_cleaning.R
```diff
@@ -211,8 +221,14 @@ qc.rbind_inprogress <- function(inprogress_qdf, clean_qdf){
...
   # responses are closed and set to complete, and any survey elements that
   # depend on them (e.g, survey flow) should still work properly.
-
-  # First, just stop if the in-progress qdf and clean qdf have different
+  
+  # First, Qualtrics in-progress datasets automatically append an extra 
+  # column containing information about progress. This will break 
+  # rbinding, so append such a column to the clean_qdf.
+  
+  clean_qdf$partial_response_progress <- NA
```
chris-perts: So other scripts were taking care of this lack-of-progress-column problem themselves before, right? I want to make sure this code gracefully handles that. The only problem case I can see is if those other scripts put information (not just NA) in there which you are overwriting with NA here. But I assume you guys never do that.

R/util_scale_computation.R
```diff
@@ -18,7 +18,7 @@ sc.append_scales <- function(
     # filter any non-scale variables from scale_variables_table
     scale_variables_table <- filter(
         scale_variables_table,
-        !is.na(scale)
+        !util.is_blank(scale)
```
chris-perts: Just to double check: are there never cases where a "blank" value, which would be newly filtered by `util.is_blank()`, would count as information you _don't_ want to filter? Like an empty string?

R/util_data_summaries.R
sarahgripshover: The two notations are _usually_ equivalent, even though I agree that your interpretation makes more sense. I've never figured out the exact conditions under which the former notation "breaks" (i.e., exhibits behavior similar to what you describe—but instead of returning a one-column df, it returns a one-column matrix, which is just no good for anybody. I suspect it depends on which version of R you are running, though I could be wrong). But I've noticed that sometimes it does break, whereas the latter notation exhibits consistent behavior.

R/util_scale_computation.R
sarahgripshover: I can't imagine there being empty string values, spaces, etc. that we'd want to retain in this context.

R/util_qualtrics_cleaning.R
sarahgripshover: I'm actually not sure what they were doing before. It would probably be a good idea to check on this.

chris-perts:
@sarahgripshover you can mash that sweet "merge" button if you still need this code.

Otherwise, please comment that it's not longer useful and close the PR.

PR #31, opened 2018-01-12
Util df compare

dkgreene:
Hi guys,

This is not an urgent pull-request, but you may find the code here useful for quickly comparing data frames.

This is a set of proposed Gymnast extension functions for comparing two data frames. The workhorse function is util_dfc.compare_dfs, which takes in two DFs and a vector denoting if any columns in the DFs are identifiers. The function reports on the dimensions of the data frames, shared and unique column names, and shared and unique identifiers. It then subsets both data frames by the row-identifiers and column names that are shared across both data frames, taking the possibility of duplicate identifiers into account. Finally, it compares these subsetted data frames value-by-value and reports on the percentages of exact value-matches, column by column.

The canonical use-case for this function is comparing different versions of the Copilot data. I plan to use it immediately when I revise the race/ethnicity question.

This pull-request an initial draft that contains the functions themselves and some test cases. I'm new to building tools like this, and relatively new to making pull-requests, so please feel free to kick this back to me to request more work of some kind.

Best,
Dan

### DIFF
--- a/R/util_dfc.R
+++ b/R/util_dfc.R
@@ -121,9 +121,14 @@ util_dfc.compare_dfs <- function(df1, df2, id_cols = c()) {
   }
 
   # Compare dimensions
-  util.print_pre("COMPARING DIMENSIONS OF DATA FRAMES: ")
-  util.print_pre("Data frame 1: " %+% paste(dim(df1), collapse = " x "))
-  util.print_pre("Data frame 2: " %+% paste(dim(df2), collapse = " x "))
+  if(identical(dim(df1), dim(df2))) {
+    util.print_pre("Data frames have the same dimensions.")
+  } else {
+    util.print_pre("Data frames have different dimensions!")
+    util.print_pre("Data frame 1: " %+% paste(dim(df1), collapse = " x "))
+    util.print_pre("Data frame 2: " %+% paste(dim(df2), collapse = " x "))
+  }
+
 
 
   ##### COMPARING AND FILTERING COLUMNS
--- a/R/util_dfc.R
+++ b/R/util_dfc.R
@@ -136,26 +136,33 @@ util_dfc.compare_dfs <- function(df1, df2, id_cols = c()) {
   # Compare column names
   util.print_pre("COMPARING COLUMN NAMES: ")
   col_name_compare_list <- util_dfc.compare_colnames(df1, df2)
-  util.print_pre("Shared column names: " %+%
-                   paste(col_name_compare_list$shared_elements, collapse = ", "))
-  util.print_pre("Columns only in first data frame: " %+%
-                   paste(col_name_compare_list$only_in_first, collapse = ", "))
-  util.print_pre("Columns only in second data frame: " %+%
-                   paste(col_name_compare_list$only_in_second, collapse = ", "))
+  col_names_not_perfect_match <- ifelse(length(c(col_name_compare_list$only_in_first, col_name_compare_list$only_in_second)) > 0,
+                                        T, F)
+  # If the column names don't match up perfectly...
+  if(col_names_not_perfect_match) {
 
-  # Subset to shared columns if needed.
-  if(length(c(col_name_compare_list$only_in_first, col_name_compare_list$only_in_second)) > 0) {
+    # Report the column names that didn't match
+    util.print_pre("Column names did not match perfectly.")
+    util.print_pre("Columns only in first data frame: " %+%
+                     paste(col_name_compare_list$only_in_first, collapse = ", "))
+    util.print_pre("Columns only in second data frame: " %+%
+                     paste(col_name_compare_list$only_in_second, collapse = ", "))
+
     # If there are NO matching columns, exit!
     if(length(col_name_compare_list$shared_elements) == 0) {
       util.print_pre("Data frames share no common column names! Comparison is impossible. Exiting.")
       return()
     }
+
     # otherwise, subset to shared columns
-    util.print_pre("Column names don't perfectly match, so only comparing shared columns: " %+%
-                     paste(col_name_compare_list$shared_elements, collapse = ", "))
+    util.print_pre("Only comparing shared columns going forward.")
     df1 <- df1[, col_name_compare_list$shared_elements]
     df2 <- df2[, col_name_compare_list$shared_elements]
+
+  } else {
+    util.print_pre("Column names matched perfectly!")
   }
+
 
   # Sort columns for apples-to-apples comparison.
   df1 <- df1[, sort(names(df1))]
--- a/R/util_dfc.R
+++ b/R/util_dfc.R
@@ -19,12 +19,10 @@ dfc.setdiff_plus <- function(vec1, vec2) {
   # First, warn if there are duplicates in either vector. They will be discarded in the setdiff.
   # Warn if there are duplicate IDs in either df!
   if(any(duplicated(vec1))) {
-    warning("Warning - first argument passed to dfc.setdiff_plus has duplicate values that will be ignored in the set-diff:")
-    message(unique(vec1[duplicated(vec1)]))
+    message("Warning - first argument passed to dfc.setdiff_plus has duplicate values that will be ignored in the set-diff.")
   }
   if(any(duplicated(vec2))) {
-    warning("Warning - second argument passed to dfc.setdiff_plus has duplicate values that will be ignored in the set-diff:")
-    message(unique(vec2[duplicated(vec2)]))
+    message("Warning - second argument passed to dfc.setdiff_plus has duplicate values that will be ignored in the set-diff.")
   }
 
   return(list(shared_elements = intersect(vec1, vec2),
@@ -44,24 +42,9 @@ dfc.get_concatenated_ids <- function(df, id_cols) {
 }
 
 
-
-
-
-
-
-
-
-
-
-
-
-
-
-
-
 dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
   # Function for comparing exact values of two data frames that have identical columns and row identifiers.
-  # INPUTS: two data frames and a string naming a shared ID column.
+  # INPUTS: two data frames and a string naming a shared ID column. "Verbose" prints heads of summary tables.
   # OUTPUT: A list containing three elements:
   #    diff_df: a DF the same dimensions as the input DF of TRUEs and FALSEs for value comparisons.
   #    row_summary: summary table of percent-matched across row identifiers.
@@ -98,7 +81,7 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
 
   # warn if there are duplicates in the ID column for either, because they make precise row sorting impossible
   if(any(duplicated(df1[, id_col])) | any(duplicated(df2[, id_col]))) {
-    warning("Warning - duplicate ID values found in at least one data frame. Rows might not be sorted the same way, even if row IDs are.")
+    message("Warning - duplicate ID values found in at least one data frame. Rows might not be sorted the same way, even if row IDs are.")
   }
 
   ##### ALL CHECKS PASSED - BEGIN VALUE-BY-VALUE COMPARISON #####
--- a/R/util_dfc.R
+++ b/R/util_dfc.R
@@ -14,8 +14,6 @@
 setwd("~/Sites/gymnast/R")
 source("util.R")
 
-
-
 util_dfc.compare_vecs <- function(vec1, vec2) {
   # compare vector elements without regard to order.
   # return a list with three components -
@@ -27,6 +25,16 @@ util_dfc.compare_vecs <- function(vec1, vec2) {
               only_in_second = setdiff(vec2, vec1)))
 }
 
+
+util_dfc.get_condensed_ids <- function(df, id_cols) {
+  # Helper function for getting a vector of IDs (not necessarily unique) from a DF.
+  # If there are multiple ID columns, they need to be concatenated with "__".
+  if(length(id_cols) == 1) {
+    return(as.character(df[, id_cols]))
+  } else {
+    return(apply(df[, id_cols], 1, paste, collapse = "__"))
+  }
+}
 
 
 util_dfc.compare_colnames <- function(df1, df2) {
@@ -50,8 +58,9 @@ util_dfc.compare_colnames <- function(df1, df2) {
 util_dfc.compare_identifiers <- function(df1, df2, id_cols, id_cols_uniquely_identify_rows = FALSE) {
   # Compare two dfs on a set of one or more commonly-shared columns that identify rows.
   # If rows are identified by multiple columns (e.g. team x student x week), include all column names in id_cols.
+  # Return a list with shared IDs, IDs in df1 only, and IDs in df2 only.
   #
-  # If you expect id_cols to UNIQUELY identify rows in both dfs, set id_cols_uniquely_identify_rows to TRUE.
+  # If you expect id_cols to UNIQUELY identify each row in both dfs, set id_cols_uniquely_identify_rows to TRUE.
   # Otherwise, duplicate ID values within each data frame will get ignored.
   # Example use case: you want to compare a student-x-week df with a student df to see if they have the same students.
   # The commonly-shared id_col is student ID.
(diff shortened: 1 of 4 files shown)


R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # Also cut the temp_until_id column as you do the comparison.
+  dfdiff <- as.data.frame(df1[, !names(df1) %in% "temp_util_id"] == df2[, !names(df2) %in% "temp_util_id"])
+  ds.summarize_by_column(dfdiff, func_list = list("pct_matched" = mean)) %>%
+                           util.html_table()
+  if(all(dfdiff == TRUE)) {
+    util.print_pre("After subsetting and sorting on shared columns (by name) and rows (by ID, if possible), the data frames match.")
+  }
+
+}
+
```
rumenPERTS: I expected two additional pieces of information:
1. List (and count) of column names for which the values are not identical, and how many non-matching values. For example, the output can be something like column A (12 non-matches), column C (2 non-matches), etc.... If more than 5 columns differ, you can truncate them.
2. List (and count) of ids which do not match. For example studen_a_week_3 (4 non-matches), student_b_week_1 (2 non_matches). Again, truncate if more than 5.

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  ## shared_elements: the shared elements,
+  ## only_in_first: the elements in vec1 and not vec2,
+  ## only_in_second: the elements in vec2 and not vec1.
+  return(list(shared_elements = intersect(vec1, vec2),
+              only_in_first = setdiff(vec1, vec2),
+              only_in_second = setdiff(vec2, vec1)))
+}
+
+
+util_dfc.get_condensed_ids <- function(df, id_cols) {
```
sarahgripshover: I might call this function `get_concatenated_ids`...they're not really being shortened

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+###     util_dfc.R
+###     Dan Greene, 12/21/17
+###
+###     The goal of the util_dfc ("data frame compare") module is to make it easy to see
+###     where two data frames differ for debugging and further analysis.
+###
+###     Depends on util.R
+###
+
+util_dfc.compare_vecs <- function(vec1, vec2) {
```
sarahgripshover: I'd make the module prefix here a little simpler, maybe just `dfc` (our scale computation modules are just `sc.function` and our graphics functions are just `ug.function`, data summarise functions are `ds.function` etc.)

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # compare vector elements without regard to order and without checking for duplicates.
+  # return a list with three components -
+  ## shared_elements: the shared elements,
+  ## only_in_first: the elements in vec1 and not vec2,
+  ## only_in_second: the elements in vec2 and not vec1.
+  return(list(shared_elements = intersect(vec1, vec2),
+              only_in_first = setdiff(vec1, vec2),
+              only_in_second = setdiff(vec2, vec1)))
```
sarahgripshover: I really like this only_in_first, only_in_second convention that you use throughout.

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+
+
+
+util_dfc.compare_identifiers <- function(df1, df2, id_cols, id_cols_uniquely_identify_rows = FALSE) {
+  # Compare two dfs on a set of one or more commonly-shared columns that identify rows.
+  # If rows are identified by multiple columns (e.g. team x student x week), include all column names in id_cols.
+  # Return a list with shared IDs, IDs in df1 only, and IDs in df2 only.
+  #
+  # If you expect id_cols to UNIQUELY identify each row in both dfs, set id_cols_uniquely_identify_rows to TRUE.
+  # Otherwise, duplicate ID values within each data frame will get ignored.
```
sarahgripshover: I'm not sure I understand this comment. Ignored, how? What is the opposite of ignoring? Is the opposite to include duplicates as un-shared identifiers if they are duplicated in one df but not the other?

If so, then it seems to me like this is conceptually different from evaluating whether the identifiers in one df are contained in the other, and it might make sense to separate that into a separate function.

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+
+  # Sanity checks
+  if(!is.data.frame(df1) | !is.data.frame(df2)) {
+    stop("Error - at least one of the first two arguments for util_dfc.compare_unique_identifiers is not a data frame.")
+  }
+  if(is.na(id_cols) | length(id_cols) == 0) {
+    stop("Error - id_cols argument is NA or empty.")
+  }
+  if(any(!id_cols %in% names(df1)) | any(!id_cols %in% names(df2))) {
+    stop("Error - not all id columns provided are in both data frames.")
```
sarahgripshover: I always make the error message say which values are missing, e.g.,

```
missing_cols <- id_cols[!id_cols %in% names(df)]
stop("Error - The following id columns provided were missing from one or more data.frames:" %+%
    paste0(missing_cols, collapse = ", ")
```

R/util_dfc.R
sarahgripshover: in fact, I have a function that checks unique indices that I copy and paste a lot—perhaps all or part could be useful here

```
check_index_is_unique <- function(df, index_cols = NULL){
    # Checks whether the vector of column names index_cols define 
    # unique rows in df by comparing the number of rows in the whole df
    # to the number of rows defined by just the unique combinations of the 
    # index_cols. Returns boolean if throw_error is FALSE, otherwise simply
    # throws an error if the check fails.
    
    # if no index cols are specified, then check whether the whole df is unique
    if(is.null(index_cols)){
        index_cols <- names(df)
    }
    
    # Throw an error if df is missing any index_cols
    if(any(!index_cols %in% names(df))){
        stop("in check_index_is_unique, one of your index cols does not exist in df.")
    }
    
    nrows_index_cols <- df %>%
        select(one_of(index_cols)) %>%
        unique %>%
        nrow
    
    nrows_whole_df <- nrow(df)
    
    return(nrows_index_cols == nrows_whole_df)
}
```

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # Are they data frames?
+  if(!is.data.frame(df1) | !is.data.frame(df2)) {
+    stop("Error - at least one of the first two arguments for util_dfc.compare_dfs is not a data frame.")
+  }
+
+  # Compare dimensions
+  util.print_pre("COMPARING DIMENSIONS OF DATA FRAMES: ")
+  util.print_pre("Data frame 1: " %+% paste(dim(df1), collapse = " x "))
+  util.print_pre("Data frame 2: " %+% paste(dim(df2), collapse = " x "))
+
```
sarahgripshover: If we use these functions in non-interactive mode, it could be annoying not to be able to turn off util.print_pre. It might make more sense here to just use `message()` or something that can be turned off in rmarkdown chunks if desired.

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # Compare column names
+  util.print_pre("COMPARING COLUMN NAMES: ")
+  col_name_compare_list <- util_dfc.compare_colnames(df1, df2)
+  util.print_pre("Shared column names: " %+%
+                   paste(col_name_compare_list$shared_elements, collapse = ", "))
+  util.print_pre("Columns only in first data frame: " %+%
+                   paste(col_name_compare_list$only_in_first, collapse = ", "))
+  util.print_pre("Columns only in second data frame: " %+%
+                   paste(col_name_compare_list$only_in_second, collapse = ", "))
+
```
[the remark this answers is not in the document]
sarahgripshover: +1

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+      dup_IDs <- TRUE
+      util.print_pre("WARNING - the second data frame has duplicate ID rows (as defined by the id_cols parameter). This makes precise row-matching impossible.")
+    }
+
+    # Report on the overlap between unique IDs across data frames
+    unique_id_compare_list <- util_dfc.compare_vecs(unique(df1$temp_util_id), unique(df2$temp_util_id))
+    util.print_pre("COMPARING IDENTIFIERS: ")
+    util.print_pre(length(unique_id_compare_list$shared_elements) %+% " shared unique identifiers, such as: " %+%
+                     paste(head(unique_id_compare_list$shared_elements), collapse = ", "))
+    util.print_pre(length(unique_id_compare_list$only_in_first) %+% " unique identifiers only in first data frame.")
```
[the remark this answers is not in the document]
sarahgripshover: +1

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+
+  # use util_dfc.compare_vecs on id_cols of each df
+  return(util_dfc.compare_vecs(df1_ids, df2_ids))
+}
+
+
+
+
+util_dfc.compare_dfs <- function(df1, df2, id_cols = c()) {
+  # Big wrapper function for giving an overall picture of how two DFs differ:
```
sarahgripshover: A high-level thought on this wrapper function: I'm not sure the gymnast package needs to have a wrapper function like this. I expect many projects to draw on these df comparison functions, not just the EP. And I think the set and order of comparison operations is likely to differ from project to project, or even from comparison to comparison within projects.

For example, I am 100% excited to use this function suite in the IES data pipeline! I think we will frequently want to ensure that although new identifiers (rows in combination) have been added, none have been subtracted. This suite offers a very elegant set of functions to do that, but I would likely use your compare_identifiers function a la carte, rather than as part of a wrapper function, for this purpose.

If we do decide we need a wrapper function, then I might suggest having it return an object including mismatched identifiers, missing column names, etc., rather than printing a report to the screen. When datasets don't match, you almost always need access to the offending identifiers, column names, etc. for the purposes of figuring out wtf went wrong (which is why I _love_ that your other functions do return objects with this information—and another part of why I'm not convinced we need a whole wrapper).

sarahgripshover:
Dan, this is great! I love these checks. I've made a few comments, and since this is for Gymnast I'd like to get @chris-perts 's eyes on it too, if possible. I suggested to him that it might make a good topic for dev chat.

You wrote a test script, which is great—if we decide to eliminate or de-emphasize the wrapper function, then it might make sense to adapt these tests to work with the other functions, rather than the wrapper function.

Looking forward to using this code ASAP!

R/util_dfc.R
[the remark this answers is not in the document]
dkgreene: I agree with your earlier point, but in this case I'm only printing the head (first 6 elements) of the shared elements. I think it will be useful for debugging identifier problems to have a few examples of identifiers that DO match as well as a few that don't.

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+
+    # Report on the overlap between unique IDs across data frames
+    unique_id_compare_list <- util_dfc.compare_vecs(unique(df1$temp_util_id), unique(df2$temp_util_id))
+    util.print_pre("COMPARING IDENTIFIERS: ")
+    util.print_pre(length(unique_id_compare_list$shared_elements) %+% " shared unique identifiers, such as: " %+%
+                     paste(head(unique_id_compare_list$shared_elements), collapse = ", "))
+    util.print_pre(length(unique_id_compare_list$only_in_first) %+% " unique identifiers only in first data frame.")
+    if(length(unique_id_compare_list$only_in_first) > 0) {
+      util.print_pre("...such as: " %+% paste(head(unique_id_compare_list$only_in_first), collapse = ", "))}
+    util.print_pre(length(unique_id_compare_list$only_in_second) %+% " unique identifiers only in second data frame.")
```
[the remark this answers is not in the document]
dkgreene: Yep I'm only printing head()

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # Sanity-check that the dimensions of the two DFs are the same
+  if(!all(dim(df1) == dim(df2))){
+    util.print_pre("WARNING - data frames do not have same dimensions after attempting to filter to shared rows and columns. No way to compare values. Stopping function.")
+    return()
+  }
+
+  # Actually compare the two data frames and tabulate the number of matches for each column.
+  # First, cast NAs to the string "NA" so that "NA" == "NA" returns TRUE.
+  df1[is.na(df1)] <- "NA"
+  df2[is.na(df2)] <- "NA"
```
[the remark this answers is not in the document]
dkgreene: Yeah that's fair. I'll change it to "__NA__".

R/util_dfc.R
dkgreene: #1 makes sense, thanks! #2 I think I do in the earlier section that starts with "COMPARING IDENTIFIERS: "

dkgreene: Actually I think I do something close to Point 1 already when I apply ds.summarize_by_column() to the data frame dfdiff. I'll add a function to count non-matches.

R/util_dfc.R
dkgreene: love it

R/util_dfc.R
dkgreene: yep

R/util_dfc.R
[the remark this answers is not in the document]
dkgreene: done!

R/util_dfc.R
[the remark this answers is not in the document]
dkgreene: yep!

R/util_dfc.R
```diff
@@ -0,0 +1,238 @@
...
+  # The commonly-shared id_col is student ID.
+  # Since student ID doesn't uniquely identify rows in the first df, id_cols_uniquely_identify_rows = FALSE.
+
+  # Sanity checks
+  if(!is.data.frame(df1) | !is.data.frame(df2)) {
+    stop("Error - at least one of the first two arguments for util_dfc.compare_unique_identifiers is not a data frame.")
+  }
+  if(is.na(id_cols) | length(id_cols) == 0) {
+    stop("Error - id_cols argument is NA or empty.")
+  }
```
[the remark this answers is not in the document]
dkgreene: great catch!

R/util_dfc.R
dkgreene: Very helpful feedback. FWIW, I have now updated this wrapper function to return two things: a summary table of which COLUMNS have mismatched values and how many, and a summary table of which ROW IDENTIFIERS have mismatched values and how many. These value-by-value comparisons are only possible if you do extensive sorting and pre-cleaning, so a wrapper function might be necessary.

R/util_dfc.R
dkgreene: Cool - switching to message().

R/util_dfc.R
dkgreene: Yeah this is very helpful and my function is not clear. I'd like to talk on Monday about how we'd like all of these functions to handle duplicates.

R/util_dfc.R
dkgreene: Yes!

dkgreene:
Notes for devchat: I'm not sure how these functions should handle duplicates in general.

* Right now dfc.compare_vecs cuts duplicates, but that's not clear behavior.
* dfc.compare_identifiers has this "id_cols_uniquely_identify_rows" argument, but its function isn't clear; maybe a clearer argument would be something like "remove_duplicates".

I think that there could be cases where you do want to compare data frames and identifier-vectors that have duplicate values, and you don't want to remove them - you want to know that the numbers of duplicates are the SAME in both DFs.

I also don't want to secretly cut duplicates, because that's not intuitively clear behavior.

So I think the ideal behavior is to give the user control over whether to cut duplicates in higher-level functions like dfc.compare_identifiers and dfc.compare_dfs, and then do the cutting before handing things over to simpler low-level functions like dfc.compare_vecs.

R/util_dfc.R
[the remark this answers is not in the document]
sarahgripshover: Ought to be careful about printing identifiers at all, though. If they got printed in an .Rmd file and synced on our repo, that would be a breach of security. I recommend not printing them at all.

R/util_dfc.R
sarahgripshover: looking forward to it! :)

sarahgripshover:
> I think that there could be cases where you do want to compare data frames and identifier-vectors that have duplicate values, and you don't want to remove them - you want to know that the numbers of duplicates are the SAME in both DFs.

I could imagine this too, but I don't think we should worry about building every possible functionality that we may need. There are so many checks we could possibly create that I think we should let our immediate needs guide us. Do we need to run this duplicates-comparison in the next 3 months? If yes, then let's build it. If not, then let's keep the set-comparisons for identifiers and not worry about duplicates.

R/util_dfc.R
```diff
@@ -0,0 +1,283 @@
...
+###     util_dfc.R
+###     Dan Greene, 12/21/17
+###
+###     The goal of the util_dfc ("data frame compare") module is to make it easy to see
+###     where two data frames differ for debugging and further analysis.
+###
+###     Depends on util.R and util_data_summaries.R
+###
+
+dfc.compare_vecs <- function(vec1, vec2) {
```
sarahgripshover: change function name to something like `dfc.setdiff_plus`

PR #35, opened 2018-03-05
added fread functions

sarahgripshover:
# Overview

I added two functions to incorporate the function `fread` into our file-reading scheme. The first is `custom_fread`, which (1) sets some basic parameters to `fread`; (2) prevents `fread` from trying to guess the type of data in each column, and instead reads columns in as character types; and (3) re-assigns types using util functions (e.g., `util.as_numeric_if_number`)

Note that I also had to include a couple of extra functions for converting booleans back to logical types. Unit tests for these new functions are also included.

# Testing

**No systematic testing has been done on the `custom_fread` function or its counterpart, `__util.read_csv_files_fast`**, because we don't have the bandwidth on our team to test it sufficiently at the moment. In the meantime, the analysts will test it "as we go," logging instances where its output differs from that of `util.read_csv_files`.

I _did_, however, test the conversion-to-logicals functions, and unit tests are included in the PR

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
     return(x)
 }
 
+util.is_vector_of_logicals <- function(x){
+  # Are all elements of x logicals (permitting blanks, NOT permitting vectors entirely 
+  # composed of numeric booleans, i.e., 0/1)?
+  # (regardless of whether x is numeric)
+  # Character vectors sometimes need to be changed to logicals,
+  # e.g., when all columns set to character types by default.
+  
+  # if all the elements of x are numbers, return FALSE. This is to ensure correct handling of
+  # binary 0/1 variables as numbers, not logicals
+  if(util.is_vector_of_numbers(x)){
+    return(FALSE)
+  }
+  
+  # find logical values
+  x_as_logical <- suppressWarnings(as.logical(x))
+  
+  # anything that gets coerced to NA by as.numeric is not really a number.
+  # but elements that were blank originally are ok.
+  non_logical <- is.na(x_as_logical)
+  originally_blank <- util.is_blank(x)
+  
+  # Return FALSE if there are any coerced blanks, because that means x 
+  # contained elements that could not be converted to numeric.
+  # Otherwise return TRUE.
+  if(any(non_logical & !originally_blank)){
+    return(FALSE)
+  }
+  else{
+    return(TRUE)
+  }
+}
+
+util.is_vector_of_logicals_test <- function(){
+  x_char <- c("TRUE", "FALSE", "a", "b")
+  x_logical <- c("TRUE", "FALSE")
+  x_0_1_only <- c(1, 0)
+  x_logical_blanks <- c(x_logical, "NA", NA, "")
+  
+  if(is_vector_of_logicals(x_char) == TRUE){
+    stop("In is_vector_of_logicals, vectors containing " %+%
+           "character values are being evaluted as numbers.")
+  }
+  if(is_vector_of_logicals(x_logical) == FALSE){
+    stop("In is_vector_of_logicals, vectors containing " %+%
+           "only 'TRUE' and 'FALSE' string values are being evaluted as non-logicals.")
+  }
+  if(is_vector_of_logicals(x_0_1_only) == TRUE){
+    stop("In is_vector_of_logicals, vectors containing " %+%
+           "only 0/1 values are being evaluated as logical. They " %+%
+           "should not evaluate as logical, because it is more likely that " %+%
+           "they are vectors of numbers.")
+  }
+  if(is_vector_of_logicals(x_logical_blanks) == FALSE){
+    stop("In util.is_vector_of_numbers, vectors containing " %+%
+           "only 'TRUE' and 'FALSE' string values and blanks are being evaluated as non-logicals.")
+  }
+}
+
+is_vector_of_logicals_test()
+
+as_logical_if_logical <- function(x){
+  # run as.numeric if the x is made up of numbers
+  # runs independently on each column if x is data.frame
+  if(class(x) %in% "data.frame"){
+    return(util.apply_columns(x, is_vector_of_logicals))
+  }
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -706,6 +778,65 @@ util.find_crypt_paths <- function (files_to_load, initial_path = NA,
...
+    sep = ",",
+    na.strings = c("NA", ""),
+    colClasses = "character",
+    ...
+  ) %>%
+    util.to_ascii %>%
+    util.apply_columns(., util.as_numeric_if_number) %>%
+    util.apply_columns(., as_logical_if_logical)
+}
+
```
rumenPERTS: Hardcoding missing values and separation symbols reduces the flexibility the function. I'm not sure what a good solution is, since you cannot just add a test of this assumption is correct. May be have these as default parameters?

sarahgripshover: yeah, good point. We could also just not include the `na.strings = c("NA", "")` parameter at all, and maybe change the function name to `util.fread_csv` so that the `sep = ","` makes more sense...the `colClasses` one is intrinsic to the function's purpose, so that one would need to stay.

R/util.R
```diff
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
...
+  # if all the elements of x are numbers, return FALSE. This is to ensure correct handling of
+  # binary 0/1 variables as numbers, not logicals
+  if(util.is_vector_of_numbers(x)){
+    return(FALSE)
+  }
+  
+  # find logical values
+  x_as_logical <- suppressWarnings(as.logical(x))
+  
+  # anything that gets coerced to NA by as.numeric is not really a number.
```
dkgreene: Nit: don't you mean "anything that gets coerced to NA by as.logical is not really a boolean"?

R/util.R
```diff
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
...
+  # but elements that were blank originally are ok.
+  non_logical <- is.na(x_as_logical)
+  originally_blank <- util.is_blank(x)
+  
+  # Return FALSE if there are any coerced blanks, because that means x 
+  # contained elements that could not be converted to numeric.
```
dkgreene: "elements that could not be converted to BOOLEAN"?

R/util.R
```diff
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
...
+  if(is_vector_of_logicals(x_logical_blanks) == FALSE){
+    stop("In util.is_vector_of_numbers, vectors containing " %+%
+           "only 'TRUE' and 'FALSE' string values and blanks are being evaluated as non-logicals.")
+  }
+}
+
+is_vector_of_logicals_test()
+
+as_logical_if_logical <- function(x){
+  # run as.numeric if the x is made up of numbers
```
dkgreene: "run as.logical if the x is made up of booleans"

R/util.R
dkgreene: +1 to Sarah's idea

R/util.R
```diff
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
...
+  if(is_vector_of_logicals(x_0_1_only) == TRUE){
+    stop("In is_vector_of_logicals, vectors containing " %+%
+           "only 0/1 values are being evaluated as logical. They " %+%
+           "should not evaluate as logical, because it is more likely that " %+%
+           "they are vectors of numbers.")
+  }
+  if(is_vector_of_logicals(x_logical_blanks) == FALSE){
+    stop("In util.is_vector_of_numbers, vectors containing " %+%
+           "only 'TRUE' and 'FALSE' string values and blanks are being evaluated as non-logicals.")
+  }
```
chris-perts: Assuming you agreed with my earlier comments on how to write tests and asserts, let's make them consistent, e.g.

```
assert <- stopifnot

assert(!is_vector_of_logicals(x_char))
assert(!is_vector_of_logicals(x_0_1_only))
assert(is_vector_of_logicals(x_logical))
assert(is_vector_of_logicals(x_logical_blanks))
```

R/util.R
```diff
@@ -184,6 +184,78 @@ util.as_numeric_if_number <- function(x){
...
+  else{
+    return(TRUE)
+  }
+}
+
+util.is_vector_of_logicals_test <- function(){
+  x_char <- c("TRUE", "FALSE", "a", "b")
+  x_logical <- c("TRUE", "FALSE")
+  x_0_1_only <- c(1, 0)
```
chris-perts: nit: I'd expect a broad range of numbers to also pass the "coerces to logical without becoming NA" test, which seems like something you want to protect against. So how about adding a case like

```
x_numbers <- c(-1, 0, 0.1, 1, 2)  # should not be detected as vector of logicals
```

That said, if the tests for `util.is_vector_of_numbers` already cover these cases, no need to repeat them here.

R/util.R
```diff
@@ -706,6 +778,65 @@ util.find_crypt_paths <- function (files_to_load, initial_path = NA,
 }
 
 
+util.custom_fread <- function(path, ...){
+  # setting colClasses to character prevents non-lossless coersion of column types by fread;
```
chris-perts: nit: A more typical word for "non-lossless" is "lossy".

R/util.R
```diff
@@ -706,6 +778,65 @@ util.find_crypt_paths <- function (files_to_load, initial_path = NA,
...
+    data.table = FALSE,
+    sep = ",",
+    na.strings = c("NA", ""),
+    colClasses = "character",
+    ...
+  ) %>%
+    util.to_ascii %>%
+    util.apply_columns(., util.as_numeric_if_number) %>%
+    util.apply_columns(., as_logical_if_logical)
+}
```
chris-perts: I would love to see the ability to specify the types I want to see for certain columns as an option. For instance, what if I could write:

```
col_types <- list(
  gpa = double,
  graduated = logical
)
util.fread_csv('my.csv', col_types)
```

Any columns not mentioned could get the default behavior you already have, but then you could also be sure that certain columns got treated as numbers even if the data was dirty.

PR #36, opened 2018-03-05
Dfc add side by side

sarahgripshover:
# Overview

I'm submitting the supplementary functions I used to employ `util.dfc` into the IES data pipeline. Functions do the following:

* Function to clean meaningless differences before comparison (e.g., blanks, rounding errors, special characters)
* Wrapper function to run index, column, and values comparison and return a "comparison" object used by other functions
* Addition to `dfc.compare_df_values` to return a side-by-side comparison of values for non-matching columns (in addition to the `dfdiff`)
* Wrapper function to make a quick summary in data.frame form of results of comparison operations
* Function to return a summary of all differing values, in easy format to summarize with dplyr functions

# Testing/Feedback
The only testing I've done is to use the functions in my own pipeline analysis. Until we have time for more systematic testing, these functions will remain in this branch. The next person/people to use them ought to review the functions, see if they duplicate other efforts out there (i.e., if there are R packages that accomplish this same purpose), and if they don't, think about how the functions are organized, what revisions should be made, and what testing ought to be done.

### DIFF
--- a/R/util_dfc.R
+++ b/R/util_dfc.R
@@ -10,11 +10,19 @@
 ###
 
 dfc.setdiff_plus <- function(vec1, vec2) {
-  # compare unique vector elements without regard to order and after automatically discarding any duplicates.
-  # return a list with three components -
-  ## shared_elements: the shared unique elements,
-  ## only_in_first: the unique elements in vec1 and not vec2,
-  ## only_in_second: the unique elements in vec2 and not vec1.
+  # Compare unique vector elements without regard to order and after
+  # automatically discarding any duplicates.
+  #
+  # Args: two vectors to compare
+  #
+  # Returns: list with three components:
+  # * shared_elements: the shared unique elements,
+  # * only_in_first: the unique elements in vec1 and not vec2,
+  # * only_in_second: the unique elements in vec2 and not vec1.
+  #
+  # Example:
+  #   dfc.setdiff_plus(c(1,2), c(2, 3))
+
 
   # First, warn if there are duplicates in either vector. They will be discarded in the setdiff.
   # Warn if there are duplicate IDs in either df!
@@ -37,10 +45,10 @@ dfc.get_concatenated_ids <- function(df, id_cols) {
   if(length(id_cols) == 1) {
     return(as.character(df[[id_cols]]))
   } else {
-    # use util.to_character do avoid some nasty behavior with apply(), which 
-    # converts numeric types to varchars when passing to paste, so that white spaces 
+    # use util.to_character do avoid some nasty behavior with apply(), which
+    # converts numeric types to varchars when passing to paste, so that white spaces
     # are appended to numbers with fewer than the max digits (e.g., if a column
-    # has both "9" and "10", the apply function turns them to " 9" and "10". 
+    # has both "9" and "10", the apply function turns them to " 9" and "10".
     # util.to_character prevents this.)
     return(apply(util.to_character(df[, id_cols]), 1, paste, collapse = "___"))
   }
@@ -118,11 +126,11 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
   # need to attach the original ids too.
   dfdiff_sum_rows[, id_col] <- df1[, id_col]
   dfdiff_sum_rows <- dfdiff_sum_rows[, c(id_col, "pct_unmatched", "num_unmatched")]
-  
+
   # get a side-by-side comparison of any unmatched values
   unmatched_summary <- dfdiff_sum_cols[dfdiff_sum_cols$num_unmatched > 0, ]
   unmatched_cols <- unmatched_summary$variable_name
-  
+
   if(length(unmatched_cols) > 0){
     side_by_side_df <- data.frame(matrix(nrow = nrow(dfdiff_sum_rows), ncol = 0))
     # we know they're all sorted the same way, so we can build the data.frame just by concatenating columns
@@ -135,7 +143,7 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
   } else{
     side_by_side_df <- NA
   }
-  
+
   if(verbose) {util.html_table(head(dfdiff_sum_rows))}
 
   # Return diff DF and full summary dfs for user.
@@ -155,7 +163,7 @@ dfc.strip_for_comparison <- function(DF){
       util.apply_columns(function(x) ifelse(util.is_blank(x), NA, x)) %>%
       util.apply_columns(function(x) if(all(is.na(x))) return(rep(NA, length(x))) else return(x))
   }
-  
+
   DF %>%
     clean_blanks %>%
     util.apply_columns(., util.as_numeric_if_number) %>%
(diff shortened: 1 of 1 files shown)


R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
...
+      n_indices_only_in_current,
+      n_indices_only_in_previous,
+      nrows_diff,
+      ncols_diff,
+      values_match
+    )
+  )
+}
+
+dfc.get_granular_mismatches <- function(values_comparison, id_cols = "index"){
```
dkgreene: This function is super awesome and worked great in my testing. But see my above comment about "id_cols" - I worry that it will be confusing to have all these nicely interoperable functions that unexpectedly make different assumptions about the id columns. I prefer requiring the user to supply one reliable ID column with the same name for both data frames, and identifying it in functions as the argument "id_col".

R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
   # need to attach the original ids too.
   dfdiff_sum_rows[, id_col] <- df1[, id_col]
   dfdiff_sum_rows <- dfdiff_sum_rows[, c(id_col, "pct_unmatched", "num_unmatched")]
-
+  
+  # get a side-by-side comparison of any unmatched values
+  unmatched_summary <- dfdiff_sum_cols[dfdiff_sum_cols$num_unmatched > 0, ]
```
chris-perts: nit: we ❤️ dplyr, so why not use `filter`?

```
unmatched_summary <- filter(dfdiff_sum_cols, num_unmatched > 0)
```

R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
...
+    for(col in unmatched_cols){
+      side_by_side_df[[col %+% "_matches"]] <- dfdiff[[col]]
+      side_by_side_df[[col %+% "_df1"]] <- df1[[col]]
+      side_by_side_df[[col %+% "_df2"]] <- df2[[col]]
+    }
+  } else{
+    side_by_side_df <- NA
+  }
+  
   if(verbose) {util.html_table(head(dfdiff_sum_rows))}
```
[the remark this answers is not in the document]
chris-perts: I'd call that a please-fix; it's pretty confusing in this location.

R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
...
+}
+
+
+dfc.strip_for_comparison <- function(DF){
+  # formats a DF to eliminate frequent causes of unimportant mismatches, including
+  # differently-formatted blanks, rounding errors, and special characters
+  clean_blanks <- function(DF){
+    DF %>%
+      util.apply_columns(function(x) ifelse(util.is_blank(x), NA, x)) %>%
+      util.apply_columns(function(x) if(all(is.na(x))) return(rep(NA, length(x))) else return(x))
```
chris-perts: please-fix-or-comment: this line (156) seems like a noop ("non-operation" i.e. doesn't do anything).

Wouldn't any `x` that satisfies `all(is.na(x))` be identical to `rep(NA, length(x))`?

R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
...
+  
+  DF %>%
+    clean_blanks %>%
+    util.apply_columns(., util.as_numeric_if_number) %>%
+    util.round_df(., 5) %>%
+    util.apply_columns(util.strip_non_ascii) %>%
+    util.apply_columns(util.strip_special_characters)
+}
 
+compare_to_previous <- function(current_df, previous_df, index_cols, compare_values = FALSE){
```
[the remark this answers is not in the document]
chris-perts: Agree with the consistency around id columns.

I don't like the name. By "compare to previous" you mean the same thing as "compare the first df you pass in with the second" which means the same thing as "data frame compare" which is the name of the whole module. Name the function after what it's doing differently from the other functions in the module. I'm not sure I can clearly delineate all the ways these functions are doing comparisons and summarizing at different levels. Spending time thinking about how to name them will help people understand that.

R/util_dfc.R
sarahgripshover: Yeah, I agree with this point. My preference would be to work with vectors of id column names throughout because that's just more similar to what I do in the rest of my R-life.

R/util_dfc.R
```diff
@@ -118,13 +118,178 @@ dfc.compare_df_values <- function(df1, df2, id_col, verbose = FALSE) {
...
+  previous_df$index <- dfc.get_concatenated_ids(previous_df, index_cols)
+  index_comparison <- dfc.setdiff_plus(current_df$index, previous_df$index)
+  indices_match <- length(index_comparison$only_in_first) == 0 & length(index_comparison$only_in_second) == 0
+  
+  # if everything else matches, and a values comparison was requested, go ahead and do a values comparison
+  if(nrows_match & ncols_match & colnames_match & indices_match & compare_values){
+    current_df <- current_df[order(current_df$index), ]
+    previous_df <- previous_df[order(previous_df$index), ]
+    values_comparison <- dfc.compare_df_values(current_df, previous_df, "index")
+    values_comparison$values_match <- all(values_comparison$col_summary$num_unmatched == 0)
```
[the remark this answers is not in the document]
sarahgripshover: yeah, come to think of it, it's pretty weird to keep this part in the wrapper function, since it modifies the `values_comparison` object that is the output of that function.

chris-perts:
We did some code review on this and came up with some idea for improvements. They're mostly half baked. We agreed @dkgreene _might_ be able to take this and run with it in April.

# Dev Chat on `dfc` organizations

To return to in April!

Adopt semi-formal "types" where we can expect a list or df has certain properties.

* <Comparison>
* ...?

## Proposed interface changes

|           current name          |          args         |                     returns                      |        new name       |
|---------------------------------|-----------------------|--------------------------------------------------|-----------------------|
| dfc.setdiff_plus                | v1, v2                | list (shared, 1st, 2nd)                          |                       |
| dfc.compare_df_values           | df1, df2, id_col      | list (row_summary, col_summary, side_by_side_df) | eqsize_compare_values |
| dfc.compare_to_previous         | df1, df2, id_cols     | <Comparison> (with summary)                      | dfc.compare_dfs       |
| dfc.compare_to_previous_summary |                       |                                                  | (dropped)             |
| dfc.get_concatenated_ids        |                       | character                                        |                       |
| dfc.get_granular_mismatches     | <Comparison>, id_cols | df of mismatches                                 |                       |
| dfc.strip_for_comparison        |                       | df same dim as input                             |                       |

## Strategy Trade-offs

Low mental overhead for users, vs. being able to cut out computationally intensive processes.

use lots of switches with defaults?

```
dfc.compare_dfs(df1, df2, id_cols, compare_values=T, summarize=T, granular=T)
```

PR #37, opened 2018-03-30
added is_present and support for matrix html tables

daveponet:
I added a new function, util.is_present, which simply returns the opposite of util.is_blank. I found it useful when trying to figure out how many values existed, where it would have been clunkier to write !util.is_blank each time.
I also added support for printing matrixes to the util.html_table function.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -107,6 +107,9 @@ util.is_blank <- function(x){
     is.na(x) | grepl("^[ \t]*$", x)
 }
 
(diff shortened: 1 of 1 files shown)


sarahgripshover:
I think you maybe didn't sync the changes to util.html_table — I can't see them in the diff

daveponet:
It's in line 309.

sarahgripshover:
oh, a very small change!

PR #39, opened 2018-06-29
add ensure_packages, use in gymnast_install()

chris-perts:
I made the function `ensure_packages` to address the problem of making sure certain packages were available, regardless of your R installation (including RServe's environment). But it's hard to use because I can't reliably _find_ that function, i.e. `source('common/ensure_packages.R')` only works if the working directory has a certain value. I can set it up to work from inside RServe, but then other analysis code doesn't work.

My new plan is to refer to this with a raw github URL, which is absolute and works regardless of the working directory:

```
source(
  "https://raw.githubusercontent.com/PERTS/gymnast/master/R/ensure_packages.R",
  local = environment()
)
```

This is the way we use gymnast already, it's just pulling a single function instead of all of `util`.

## Testing

I pulled gymnast's `util.R` from a raw github link with this change, and the `gymnast_install` function operated without complaint, because I had all the packages already. I uninstalled a package and did the same, and it installed it back for me. All packages were in my local environment via the `library()` calls, e.g. `summarise` at the command line showed the function.

It also allows common scripts like big pipe to work from other working directories, i.e. it solves the original problem:

```
setwd('~/Sites/analysis')
source('common/big_pipe.R')  # this uses ensure_packages()
# no error
```

### DIFF
--- a/R/ensure_packages.R
+++ b/R/ensure_packages.R
@@ -1,0 +1,29 @@
+ensure_packages <- function (libs, prefer_type = "binary") {
+  # Install `libs` if they aren't installed already.
+  #
+  # Args:
+  #   libs - character, packages to maybe install
+  #   prefer_type - character, default "binary", or "source", sets
+  #     options(install.packages.check.source) which is an indirect way of
+  #     controlling whether we use binary or source installation, since the
+  #     `type` argument of install.packages() is platform-dependent and thus
+  #     unreliable.
+  if (prefer_type == "binary") {
+    # Prefer binary packages, even if the source version is farther ahead.
+    # This is default because it's so much faster to install a binary.
+    options(install.packages.check.source = "no")
+  } else if (prefer_type == "source") {
+    options(install.packages.check.source = NULL)
+  }
+
+  installed <- utils::installed.packages()[,"Package"]
+  to_install <- libs[!(libs %in% installed)]
+
+  if (length(to_install)) {
+    utils::install.packages(
+      to_install,
+      repos = 'http://cran.us.r-project.org'
+      # type = 'binary'  # doesn't work for some platforms, e.g. linux
+    )
+  }
+}
--- a/R/util.R
+++ b/R/util.R
@@ -712,6 +712,8 @@ util.find_crypt_paths <- function (files_to_load, initial_path = NA,
 ###
 ###############################################################
 
+source('ensure_packages.R')
+
 gymnast_install <- function () {
     # Load (or install) all packages used by gymnast.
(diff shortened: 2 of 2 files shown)


daveponet:
Sorry if I'm being dense. Would it be simpler to just put ensure_packages in util.R?

chris-perts:
util.R is a little strange because it actually _invokes functions_ rather than merely defining them. If I did what you're suggesting, gymnast_install() would actually run, and I'd have all those packages in my global environment. I just want the definition of a single function, thus the separate file.

chris-perts:
I completely agree. That's an open problem right now. If you figure out any way to control package versions, please let me know. We'll need to get that done soon, one way or another.

daveponet:
Sounds good to me!

dkgreene:
Hi @chris-perts - just getting caught up on this and I'm a bit confused. Currently util.R doesn't run for me because the line `source('ensure_packages.R')` returns the error `could not find function "ensure_packages"`.

I amended this line of code to be the raw github URL path that you described in your first post in this thread, and everything works. But then the line `gymnast_install()` actually runs in util.R, and all the packages actually load into my workspace. This seems to _not_ be what you want, though?

I see a couple of things in the code that can be varied for different effects:

1. The ensure_packages function can be loaded via sourcing the raw Github URL, or by simply putting the function into util.R. I like the second option because it saves on having to track another file, but it's not super important and maybe there is some other benefit to the first option.

2. The gymnast_install function can be merely defined in util.R, or it can be defined and then actually run as it is now. I'm okay with either option, as long as it's consistent and we know what we need to do to get the right packages up. If you don't want packages auto-loaded when you run util.R, then you can just do option 1 here.

dkgreene:
Also, is there a way to amend the gymnast_install function to accept a vector of dependencies as an argument, with the provided ones as defaults? That would be really useful in future situations when I need partners to have certain packages installed on their local machines, e.g. the Okonofua lab.

I think this would just be as simple as changing the start of the function to:

`gymnast_install <- function (dependencies <- c(
        "digest","dplyr","formatR","ggplot2",
        "grid","Hmisc","knitr","lme4",
        "lmerTest","psych","reshape2","scales" ,
        "stargazer","stringr","xtable"
    )) {
    # Load (or install) all packages used by gymnast.`

chris-perts:
Hey @dkgreene I agree this one is messed up. I mostly just want to **not** run `gymnast_install()` automatically when you source it. Once we get that done, then being able to explicitly call it with an optional list of packages is a great idea. See #40.

PR #42, opened 2020-02-10
Util.na omit

sarah-gripshover-PERTS:
stats::na.omit has weird behavior when called on atomic types. So I wrote a function that works for vectors, matrices and data.frames (but not lists). Tests are included.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -106,6 +106,7 @@ util.na_omit_test <- function(){
   x_matrix <- matrix(nrow = 2, ncol = 2, data = c(NA, 1, 2, 3))
   x_list <- list(a = c(NA, 1), b = c(2, 3))
   x_df <- data.frame(a = c(NA, 1), b = c(2, 3))
+  x_df_allblankcol <- data.frame(a = c(1, 2), b = c(NA, NA))
   x_tibble <- dplyr::as_tibble(x_df)
 
   assert <- stopifnot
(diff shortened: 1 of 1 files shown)


R/util.R
```diff
@@ -83,6 +83,41 @@ util.duplicated_all <- function(x) {
...
+  x_number <- c(1, 2, NA)
+  x_matrix <- matrix(nrow = 2, ncol = 2, data = c(NA, 1, 2, 3))
+  x_list <- list(a = c(NA, 1), b = c(2, 3))
+  x_df <- data.frame(a = c(NA, 1), b = c(2, 3))
+  x_tibble <- dplyr::as_tibble(x_df)
+
+  assert <- stopifnot
+  assert(util.na_omit(x_char) == c("1", "a"))
+  assert(util.na_omit(x_number) == c(1, 2))
+  # I couldn't think of how to test that it throws an error for lists
```
chris-perts: I've got several working example of `testthat` code (from the Wickhamverse) that can do that kind of thing. I can show you.

R/util.R
```diff
@@ -83,6 +83,41 @@ util.duplicated_all <- function(x) {
   return (duplicated_downward | duplicated_upward)
 }
 
+util.na_omit <- function(x){
+  # returns all values of vector, matrix or data.frame x that are not NA,
+  # but maintains the type of x (in contrast with stats::na.omit which adds a
+  # bunch of unsolicited attributes to vectors)
```
chris-perts: nit: I would add that when acting on 2D data, it throws out whole columns if those columns have any NA values; came as a slight surprise to me.

R/util.R
```diff
@@ -83,6 +83,41 @@ util.duplicated_all <- function(x) {
...
+    return(x[!is.na(x)])
+  }
+  if(is.matrix(x)){
+    return(na.omit(x))
+  }
+  if(is.data.frame(x)){
+    return(x[complete.cases(x), ])
+  } else{
+    stop("util.na_omit is for vectors, matrices and data.frames only.")
+  }
```
chris-perts: nit: you aren't using `else` consistently. In the first two cases, you're taking advantage of the fact that `return()` prevents further code execution. In this last case, you're writing `else` _even though_ it's redundant with the `return()`. I suggest you drop the last else:

```
if (is.data.frame(x)) {
  return(x[complete.cases(x), ])
}
stop("...")
```

R/util.R
sarah-gripshover-PERTS: yeah I didn't actually know about that but it would surprise me too

chris-perts:
👏

PR #43, opened 2020-03-26
Gymnast used as a submodule

chris-perts:
Implements [proposal to merge code in analysis/common and gymnast](https://www.notion.so/perts/Combining-analysis-common-and-gymnast-96bdd8789d894b5eb526ea0b2cca2354) and use it all as a submodule.

The core problems to solve are:

* Loading gymnast over the Internet via github urls is slow
* RServe needs a dedicated copy of code in `common` in order to be deployed
* Division of code between gymnast and common is badly organized, both contain similarly generic modules

This moves most of analysis/common into gymnast, following `modules`-friendly practices (no `library()`, no `source()`). It keeps the package-related boilerplate, so it's no harder to build gymnast into a proper R package than it was before. It includes instructions for use as a git submodule so both analysis and RServe can have a dedicated, version-controlled copy. It incorporates unit tests.

The one negative consequence is move `util.R` to `util_legacy.R`. Existing references to gymnast via github.com URLs will have to be changed, but that's a simple find-and-replace job in the analysis repo.

## Testing

All unit tests from common pass, see comments in those files:

```
# To run these tests:
#
# * Open a terminal window
# * Change to the gymnast directory
# * Run this command:
#   Rscript -e "testthat::auto_test('R', 'tests/testthat')"
```

I used this branch as a submodule in the new [rserve repo](https://github.com/PERTS/rserve), which I've tested deployed on App Engine an can successfully render reports.


chris-perts:
> how rserve is sourcing the new gymnast submodule

You can see on the main page of the [rserve repo](https://github.com/PERTS/rserve) that gymnast is included as a submodule at a specific commit, which just looks like a copy sitting in a directory. Then the code is referenced in [server.R](https://github.com/PERTS/rserve/blob/master/server.R#L6-L8). After that you can write `import_module("whatever_you_want_in_gymnast")`.

chris-perts:
> I assume all of our analysis script file paths will have to change now, correct?

That's what I meant about:

> The one negative consequence is move util.R to util_legacy.R. Existing references to gymnast via github.com URLs will have to be changed, but that's a simple find-and-replace job in the analysis repo.

So the new-old gymnast call is:

```
source("https://raw.githubusercontent.com/PERTS/gymnast/master/R/util_legacy.R")
```

chris-perts:
@sarah-gripshover-PERTS I'm merging this because I think you _meant_ to approve(?). But you didn't actually approve, so I'd appreciate you circling back to this and making sure I didn't miss anything.

PR #44, opened 2020-06-04
Rserve 15 association helpers

chris-perts:
## Background

Addresses https://github.com/PERTS/rserve/issues/15 subtask 1 where we need a function that can sort through network relationships and find the classrooms/codes associated with them.

## Implementation

Recursively check each network association until we arrive at associations that are orgs (not other networks) then find their classrooms.

fb82d08 is needed to run gymnast tests, because I'm doing that from the `gymnast` subdirectory instead of the rserve root directory, and the normal `.Rprofile` in rserve doesn't run there.

## Testing

Unit tests included. I covered networks-of-orgs, networks-of-networks, filtering out classrooms not associated, and producing a unique set.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,0 +1,21 @@
+options(
+  repos = "http://cran.rstudio.com/", # needed to install any packages
+  install.packages.check.source = "no" # prefer binary packages
+)
+
+if (!"modules" %in% utils::installed.packages()) {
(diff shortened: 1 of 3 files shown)


chris-perts:
This doesn't track the child relationships, which is necessary for making reports. See #45. Closing.

PR #45, opened 2020-06-22
network fns in summarize_copilot.R

chris-perts:
## Background 

Addresses https://github.com/PERTS/rserve/issues/15 by building tools that let us work with network associations.

## Implementation

* `summarize_copilot$get_classrooms_from_organization()` takes a set of ids and returns all unique org-classroom associations
* `summarize_copilot$get_classrooms_from_organization()` takes a set of ids and returns all unique network-child-classroom associations, and also child names so you don't have to know anything about whether the children are orgs or networks in order to display their names or access their ids.

Included in upcoming rserve PR

## Testing

Tests included. Let me know if you think of other important test cases to cover.

## Feedback

The rest of this module, summarize_copilot, uses prefixed column names, so I kept with that e.g. `network.uid` is a typical column name. I find this helps enormously when thinking about complicated merges, and maps nicely to syntax used in SQL for similar things. Let me know if you find this pattern too verbose.

Also let me know if the final output is what you were expecting.

Note that you can use these by providing only a single id, e.g. to get the associations of one network, or you can provide a whole vector of ids.

### DIFF
--- a/.Rprofile
+++ b/.Rprofile
@@ -1,0 +1,22 @@
+options(
+  repos = "http://cran.rstudio.com/", # needed to install any packages
+  install.packages.check.source = "no" # prefer binary packages
+)
+
+if (!"modules" %in% utils::installed.packages()) {
+  # Without the INSTALL_opts here, some environments using this .Rprofile file
+  # can get in a recursive loop. R starts a new process to compile/build
+  # packages (see [source code lines 603-604][1]), which may run this .Rprofile
+  # script again. Specifying --use-vanilla tells R to ignore this file.
+  #
+  # [1]: https://www.rdocumentation.org/packages/utils/versions/3.6.2/source
+  utils::install.packages("modules", INSTALL_opts="--use-vanilla")
+}
+
+if (!exists("import_module")) {
+  modules::use("R/bootstrap.R")$install_module_imports()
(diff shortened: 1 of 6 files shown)


chris-perts:
@sarah-gripshover-PERTS ready for re-review, the table structure from _both_ functions is now the same:

```
tibble(
    parent_id = character(),
    parent_name = character(),
    child_id = character(),
    child_name = character(),
    team.uid = character(),
    team.name = character(),
    classroom.uid = character(),
    classroom.code = character(),
  )
```

Where `get_classrooms_from_organization()` always returns `Organization` parents and `Team` children, and `get_classrooms_from_network()` always returns `Network` parents and children that are either `Network` or `Organization`.

Notice also the signature of the `get_classrooms_from_organization()` function has changed, there's a new final argument, the triton organization table. You'll need to change this in `saturn_reports.R`, which should be trivial.

sarah-gripshover-PERTS:
Great, this sounds like exactly what I had in mind, and many thanks for making it happen! I'm going to need a bit more help I think with the end-to-end framework. Adding an argument to the `get_classrooms_from_organization()` function seems easy enough, but the network report loop in the working branch currently breaks. The e2es are not running at all. Maybe we can put our heads together about that late this week or early next. In the meantime, I approve this!

PR #48, opened 2020-07-08
Break up join fns

chris-perts:
## Background

We should centralize and re-use logic having to do with wrangling the Triton/Copilot schema in `summarize_copilot`. But that means the functions there need to be generic enough to use.

## Implementation

This puts each join operation of relating team, organization, and user into a separate function, while maintaining the functionality used for the dashboard in `team_community_names()`.

Notice that the function names represent left-to-right joins. So `organization_user` is what you would expect from just `merge(organization, user)` with all the complexities worked out.

## Testing

I wrote a "default case" test for each function. Not ever edge case I could think of is covered, e.g. multiple rows when a team is supervised by multiple org admins, but I was going for speed b/c SG is actively developing with this stuff.

## Misc

I also added that error SG suggested re: forgetting to prefix your tables.

### DIFF
--- a/R/summarize_copilot.R
+++ b/R/summarize_copilot.R
@@ -552,6 +552,14 @@ get_classrooms_from_organization <- function(organization_ids,
         "a platform table with zero rows."
       )
     }
+    for (col in names(table)) {
+      if (!grepl('\\.', col)) {
+        stop(
+          "summarize_copilot$get_classrooms_from_organization() received " %+%
+          "unprefixed column names."
+        )
+      }
+    }
   }
 
   team_org_assoc <- triton.team %>%
@@ -596,9 +604,17 @@ get_classrooms_from_network <- function (network_ids,
   for (table in platform_tables) {
     if (nrow(table) == 0) {
       warning(
-        "summarize_copilot$get_classrooms_from_organization() received " %+%
+        "summarize_copilot$get_classrooms_from_network() received " %+%
         "a platform table with zero rows."
       )
+    }
+    for (col in names(table)) {
+      if (!grepl('\\.', col)) {
+        stop(
+          "summarize_copilot$get_classrooms_from_network() received " %+%
+          "unprefixed column names."
+        )
+      }
     }
   }
 
--- a/R/summarize_copilot.R
+++ b/R/summarize_copilot.R
@@ -167,6 +167,15 @@ team_cycle_class_participation <- function(tcc,
   return(tcc_ppn)
 }
 
+team_user <- function (triton.team, triton.user) {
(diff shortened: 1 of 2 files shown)


chris-perts:
> Are these functions currently used anywhere in the rserve pipeline?

Nope. They're all new so they **can't** have been used elsewhere. And the `team_community_names()` is for the dashboard over in `analysis`.

chris-perts:
> Are there any heuristics to understand how many rows we would expect to be in the data.frames coming out of these functions?

They're the same rules as the basic relationships in Copilot. So you asked about `team_user()`: teams and users are many-to-many, i.e. one team can have many members _and_ one user can be on many teams. So there's no predicting how many rows, in general. Similarly orgs and users are many-to-many, and so are teams and orgs.

Networks and orgs are many-to-many (a network may have many child orgs, and orgs may have many parent networks), and networks are many-to-many with themselves (one network can have many parent networks as well as many child networks).

Things within teams are less free form: teams and classrooms are one-to-many, teams and participants are one-to-many. Classrooms and participants are many-to-many (one roster may have many students, one student may be on many rosters).

sarah-gripshover-PERTS:
> They're the same rules as the basic relationships in Copilot. So you asked about `team_user()`: teams and users are many-to-many, i.e. one team can have many members _and_ one user can be on many teams. So there's no predicting how many rows, in general. Similarly orgs and users are many-to-many, and so are teams and orgs.
 
> Networks and orgs are many-to-many (a network may have many child orgs, and orgs may have many parent networks), and networks are many-to-many with themselves (one network can have many parent networks as well as many child networks).

> Things within teams are less free form: teams and classrooms are one-to-many, teams and participants are one-to-many. Classrooms and participants are many-to-many (one roster may have many students, one student may be on many rosters).

Ok, so it sounds like in general, these functions would return data.frames with the same number of rows as unique index combos across the two data.frames, perhaps excluding teams with no users and/or users with no teams (wherever those patterns are possible). Where the relationship is one-to-many, that reduces to being the number of unique indexes in the lowest-nested df. Is that right, more or less?

chris-perts:
> these functions would return data.frames with the same number of rows as unique index combos across the two data.frames

I think I agree with your sense of things, but this isn't technically correct. The unique index in all these teams is `uid`, and it definitely isn't the cross product of all those uids. I would say, e.g. for `team_user()`, the number of rows will be the sum of the length of the `owned_teams` arrays, filtered to the team ids found in the team table. So the number of team-user relationships for those teams (that's where the "left" aspect of the join comes in, unrelated users drop off, rather than unrelated teams dropping off).

PR #49, opened 2020-07-21
util$str_percent

chris-perts:
Utility for something we do a lot: calculate a percent, format it as a string, with a percent sign, unless it's NA or divide-by-zero, in which case show something else.

Usage in tests, e.g.:

```
util$str_percent(1, 0, '-') # result: "-"
util$str_percent(1, 2) # result: "50%"
```

### DIFF
--- a/tests/testthat/test_util.R
+++ b/tests/testthat/test_util.R
@@ -44,4 +44,20 @@ describe('str_percent', {
   it('handles negative values', {
     expect_identical(util$str_percent(-1, 2), '-50%')
   })
+
+  it('handles non-unitary vectors', {
+    numerator <- c(-1, 0, 1)
+    denominator <- c(100, 100, 0)
+    expect_identical(
+      util$str_percent(numerator, denominator),
+      c('-1%', '0%', 'N/A')
+    )
+  })
+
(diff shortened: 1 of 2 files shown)


tests/testthat/test_util.R
```diff
@@ -0,0 +1,47 @@
...
+    expect_identical(util$str_percent(1, 3), '33%')
+  })
+
+  it('handles values > 100%', {
+    expect_identical(util$str_percent(100, 1), '10000%')
+  })
+
+  it('handles negative values', {
+    expect_identical(util$str_percent(-1, 2), '-50%')
+  })
```
sarah-gripshover-PERTS, line 46: Nit: might be nice to explicitly call out what I see is the main use case for a function like this; i.e.,:

```
it('handles works in dplyr pipes`, {
    d <- data.frame(a = c(1, 2, 3), b = c(2,3,4)) %>%
      mutate(perc_ab = util$str_percent(a, b)) 
   expect_equal(d$perc_ab, c("50%", "67%", "75%))
```
Right? Is that how it's supposed to work?

chris-perts: @sarah-gripshover-PERTS 🔨

PR #54, opened 2020-09-09
Optimize summarize

chris-perts:
## Background

Addresses https://github.com/PERTS/rserve/issues/87

## Implementation

I used `profiler` calls to figure out where the slowdown was: inside `summarize_copilot$map_responses_to_cycles()`. It was doing a for loop over all provided cycles, which was very slow. I pre-filtered the cycles to eliminate iterations of the loop that were irrelevant.

## Along the way

* Added some documentation
* Removed an ugly rbind loop that isn't necessary any more
* Fixed a test
* Fixed application of extended date logic. This was naïvely referencing `cycle.extended_end_date` even though that property is not defined if there is no next cycle. The right way to do this is choose _either_  `extended_end_date`, or failing that, use `end_date`. This probably was dropping responses before :-(

## Testing

Existing unit tests of `map_responses_to_cycles()` pass. Testing a rds from a recent run showed ~12 seconds before this change and ~5 seconds afterward.

## Feedback

This chunk of code is obviously pretty core to our operations. I've probably spent over 8 hours working with the nuances. It's not very long; I'd appreciate a good hard look trying to understand the logic from first principles. My only excuse for not including more tests is that it's taken me too much time to get this far...

### DIFF
--- a/R/summarize_copilot.R
+++ b/R/summarize_copilot.R
@@ -97,15 +97,22 @@ team_cycle_class <- function(team_class, triton.cycle) {
 team_cycle_class_participation <- function(tcc,
                                            triton.cycle,
                                            triton.classroom,
-                                           neptune.participant,
-                                           neptune.participant_data) {
+                                           participant,
+                                           complete_participant_data) {
   # Args:
   #   tcc - see team_cycle_class()
   #   triton.cycle - from db, include all cycles from tcc
   #   triton.classroom - from db, include all classes from tcc
-  #   neptune.participant - from db, include all ppts from...
-  #   neptune.participant - from db, include all relevant survey responses
-  if (nrow(neptune.participant_data) == 0) {
+  #   participant - df of participant-level info to join, with column
+  #     'participant.uid'. Might be from either neptune or triton dbs.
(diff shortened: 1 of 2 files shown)


R/summarize_copilot.R
```diff
@@ -98,6 +99,12 @@ team_cycle_class_participation <- function(tcc,
                                            triton.classroom,
                                            neptune.participant,
                                            neptune.participant_data) {
+  # Args:
+  #   tcc - see team_cycle_class()
+  #   triton.cycle - from db, include all cycles from tcc
+  #   triton.classroom - from db, include all classes from tcc
+  #   neptune.participant - from db, include all ppts from...
+  #   neptune.participant - from db, include all relevant survey responses
```
sarah-gripshover-PERTS, line 107: Nit: I almost always use saturn data instead of neptune participant data. Would be nice for the argument name to reflect this possibility; e.g., call it `survey_responses` and include a note that `neptune.participant_data` can also be used in a pinch, rather than the reverse

sarah-gripshover-PERTS (approved):

PR #58, opened 2020-10-26
change saturn query to use modified, not created

chris-perts:
This makes queries more predictable since a set of rows where `modified` is less than some point in the past can't mutate. Rows can disappear from that set as they are updated, but they can't _change_.


chris-perts:
Taking approval of https://github.com/PERTS/rserve/pull/108 as implicit approval of this PR.

PR #69, opened 2021-04-03
Sql.R password fix

londeree4:
## Background

This PR addresses an issue with the sql.R gymnast functions, `create_service` and `create_neptune_service`, allowing for the appropriate call of password information. 

## Implementation

Resets `password` and `password_file_name vars`, so they are called appropriately. 
Code is also restyled using styler (to remove [second commit](https://github.com/PERTS/gymnast/commit/70432423536e814cf5d315bcb7faba45c9b2244d) can be discarded). 

## Testing

Tested by including printed output of status, and utilizing the `neptune$query` function in a separate script.

### DIFF
--- a/R/sql.R
+++ b/R/sql.R
@@ -117,7 +117,7 @@ disconnect_all <- function () {
   return(length(all_conns))
 }
 
-# This sends lots of annoying warnings about various fields geting imported
+# This sends lots of annoying warnings about various fields getting imported
 # as various types, e.g. "Unsigned INTEGER in col 8 imported as numeric".
 # We're happy with how the package handles types and never want these warnings.
 query <- function (...) {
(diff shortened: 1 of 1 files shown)


londeree4:
@chris-perts - I was having difficulty committing the changes to the submodule, so I've created the same changes in the base gymnast directory. Perhaps we can chat later on the best way to commit changes when editing the submodule!

chris-perts (commented):
Thanks, Allison!

I removed the `print()` lines we added for debugging. FYI if you find a case where you really want the script to print informative output, use the `logging` module.

I also need to update the password for triton, so I added a commit for that. I have some todos on my end before this will actually work, so I'm not approving for merger yet. But consider this in my court.

chris-perts (approved):
Ok, database settings and analysis crypt updated; I tested both the neptune and triton services. Code is good to go.

chris-perts:
@londeree4 you're the author here, and the PR has approval, so please merge unless you're aware of further problems.

londeree4:
Got it! Thanks @chris-perts!

PR #70, opened 2021-04-13
Fix network assc

sarah-gripshover-PERTS:
# Description

Rserve was missing network-org combos where the same org was a member of multiple networks. 

# Implementation

Instead of taking the unique set of child_id and classroom.uid values, we take the unique rows of the whole association table (9e709cc). This avoids throwing out rows where the child is nested under more than one parent.

# Testing

9e709cc now passes. Hooray! I will do larger scale tests to make sure it interacts well with Rserve when I update the Rserve submodule inside my working branch.


chris-perts (approved):
I added a comment. Approved.

PR #71, opened 2021-05-12
Check the uniqueness of the data before imputing

sarah-gripshover-PERTS:
# Introduction

A bug was introduced by the imputation function calling `last()` on data that may not be unique by the group_by. Doing this causes imputation behavior to shift based on the sorting of the data.frame! Bad news.

# Implementation

Imputation should be happening on clean data. If there are duplicate values of the implementation index, these should be handled prior to imputation. Why? Because different scenarios might have different inclusion criteria, for one. In Rserve, we might want to always take participants' LAST response in the participation window. But certain analyses may wish to take first responses, to take a mean of all responses, etc.

More generally, it is beyond the scope of what we mean when we say "imputation" to perform this basic data processing. Any procedure we could come up with here would undoubtedly produce surprising behavior in some circumstances. Therefore, we simply throw an error when this basic data processing has not been performed and call it a day.

+ 2e6be21 is a failing test to check for an error being thrown when the imputation index is not unique.
+ f403191 makes the above test pass by stopping if the imputation index is duplicated within the time ordinal
+ 8e6e008 deletes the `last()` operation altogether because it's no longer needed (we don't need to disambiguate rows if we already know they're unique)
+ A few more things...
    - 0cb49c0 just goes ahead and writes an informative error for any expected but missing columns. (An error was already being thrown for missing columns, but it wasn't informative.)
    - 8fedfa2 and f6bad0a were minor clean-up things

# Testing
All unit tests, including the new ones described above, now pass.

### DIFF
--- a/R/imputation.R
+++ b/R/imputation.R
@@ -71,6 +71,16 @@ impute_to_time_ordinal <- function(
     stop("The time ordinal column was not found in the response data: " %+% time_ordinal_column)
   }
 
+  index_is_duplicated <- response_data %>%
+    select(all_of(c(imputation_index, time_ordinal_column))) %>%
+    duplicated()
+
+  if(any(index_is_duplicated)){
+    stop("Your imputation_index did not specify unique values within the time_ordinal column. " %+%
+           "No imputation can be performed until your response_data object contains " %+%
+           "unique values of " %+% paste0(c(imputation_index, time_ordinal_column), collapse = ", "))
+  }
+
   # handle NSE by hard-coding `time_ordinal_column` as the values in "time_ordinal_column"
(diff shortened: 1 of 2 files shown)


chris-perts (changes requested):
Forgive my paranoia, but I'd like to see a unit test that shows that differently sorted versions of the same data frame result in the same output. I know it seems obvious now that the index is unique, but this would guard against there being _another_ bug that was triggered by sort order which this fix doesn't address.

chris-perts:
And completely agree with your choice here, saying the concern of imputation is to impute, not to clean data.

sarah-gripshover-PERTS:
@chris-perts How's this? c5abfad

chris-perts (approved):

PR #74, opened 2021-08-09
Update str pct utility

londeree4:
## Background
In some cases we may want to have string percentages that do not actually have the percent sign appended (as in the case of percentage point change). 

## Implementation
[As suggested here](https://github.com/PERTS/rserve/pull/200#discussion_r682945505)  - Added scope to `str_percent` function to allow for repression of percent sign through argument `use_percent_sign` which defaults to true.

## Testing
Added in new test to test_util.R to ensure that percent sign is inhibited when requested.

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -319,9 +319,10 @@ clean_percent <- function(numerator, denominator, blank_value = 0) {
   ifelse(is_blank, blank_value, raw_pct)
 }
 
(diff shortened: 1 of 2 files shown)


sarah-gripshover-PERTS (approved):
Looks great to me. Thanks, Allison!

chris-perts (approved):
🏋️‍♂️

PR #76, opened 2021-08-23
feat: util$datetime_to_iso_string()

chris-perts:
## Background

We use datetime formats like this: `2021-08-23T21:45:40Z` but I was having a hard time getting R to spit this out. Even `lubridate` was failing me, because it wouldn't accept my specified timezone:

```
> lubridate::format_ISO8601(Sys.time(), tz="GMT", usetz=TRUE)
[1] "2021-08-23T14:22:59-0700"
```

## Implementation

Made a util function to format this the standard PERTS way. Can format any provided datetime or return the current time.

```
> util$datetime_to_iso_string()
[1] "2021-08-23T22:09:41Z"
```

### DIFF
--- a/R/util.R
+++ b/R/util.R
@@ -564,3 +564,9 @@ find_day_of_week <- function (increment, day_abbr) {
 
 next_monday <- function() find_day_of_week(+1, "Mon")
 last_monday <- function() find_day_of_week(-1, "Mon")
+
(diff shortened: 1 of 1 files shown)


sarah-gripshover-PERTS (changes requested):
I know we're working quickly, but maybe just one test to demo how it works?

PR #83, opened 2022-01-14
Extended start date

sarah-gripshover-PERTS:
# Introduction

Implements support for extended start dates, to bring reports into line with updates to the Datateer dashboard.

# Implementation

d508338 adds an `extended_start_date` field to the `map_responses_to_cycles()` function and to accompanying test data.

# Testing

+ At commit d508338, all previous unit tests still pass, which indicates the changes have not altered previously documented and tested behavior of the function. This means, when `extended_start_date` is blank, the function still uses the value in the `start_date` field.
+ 13ee6aa proves that the function assigns priority to `extended_start_date` over `start_date` when both values are present.
+ 1784a75 proves that the `extended_start_date` field is used when `start_date` is absent.

### DIFF
--- a/R/summarize_copilot.R
+++ b/R/summarize_copilot.R
@@ -310,6 +310,11 @@ map_responses_to_cycles <- function(response_tbl,
         util$is_blank(cycle.extended_end_date),
         cycle.end_date,
         cycle.extended_end_date
+      ),
+      cycle.extended_start_date = ifelse(
+        util$is_blank(cycle.extended_start_date),
+        cycle.start_date,
+        cycle.extended_start_date
       )
     )
 
@@ -329,14 +334,14 @@ map_responses_to_cycles <- function(response_tbl,
   # considering the minimum possible set of cycles.
   potential_cycles <- cycle_extended %>%
     filter(cycle.team_id %in% unique(response_merged$classroom.team_id)) %>%
-    filter(!util$is_blank(cycle.start_date))
+    filter(!util$is_blank(cycle.extended_start_date))
 
   # Fill in values of the cycle_ordinal column as their row matches various
   # cycle dates.
   for (i in sequence(nrow(potential_cycles))) {
     this_cycle <- potential_cycles[i, ]
     in_cycle <- (
-      response_merged$created_date >= this_cycle$cycle.start_date &
+      response_merged$created_date >= this_cycle$cycle.extended_start_date &
         response_merged$created_date <= this_cycle$cycle.extended_end_date &
         response_merged$classroom.team_id %in% this_cycle$cycle.team_id
     )
--- a/R/summarize_copilot.R
+++ b/R/summarize_copilot.R
@@ -332,6 +332,16 @@ map_responses_to_cycles <- function(response_tbl,
 
   # B/c we loop over every cycle, it's a big efficiency gain to make sure we're
   # considering the minimum possible set of cycles.
+  #   This considers 3 cases:
+  #
+  # 1. This is an ordinal = 1 cycle, where extended_start_date was set by Copilot.
+  # This is kept by the filter. Even if this cycle's other dates are unset, we
+  # will want assign responses to it by default.
+  # 2. This is an ordinal > 1 cycle where start_date was set by the user. Kept by the
+  # filter because code above copied this value into extended_start_date.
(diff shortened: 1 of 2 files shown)


R/summarize_copilot.R
```diff
@@ -329,14 +334,14 @@ map_responses_to_cycles <- function(response_tbl,
   # considering the minimum possible set of cycles.
   potential_cycles <- cycle_extended %>%
     filter(cycle.team_id %in% unique(response_merged$classroom.team_id)) %>%
-    filter(!util$is_blank(cycle.start_date))
+    filter(!util$is_blank(cycle.extended_start_date))
```
chris-perts, line 337: nit: I think this filter is correct, but it could probably use a comment, e.g.

> This considers 3 cases:
> 1. This is an ordinal = 1 cycle, where `extended_start_date` was set by Copilot. This is kept by the filter. Even if this cycle's other dates are unset, we will want assign responses to it by default.
> 2. This is an ordinal > 1 cycle where `start_date` was set by the user. Kept by the filter because code above copied this value into `extended_start_date`.
> 3. This is an ordinal > 1 cycle where neither `start_date` nor `extended_start_date` were are set. These are excluded by the filter because we never want to assign responses to them.

tests/testthat/test_summarize_copilot.R
```diff
@@ -469,11 +470,11 @@ describe('map_responses_to_cycles', {
...
-    ~uid,       ~team_id,     ~start_date,  ~extended_end_date,    ~ordinal,
-    'Cycle_1',  'Team_Viper', '2020-01-01', '2020-01-14',          1,
-    'Cycle_2',  'Team_Viper', '2020-01-15', '2020-01-30',          2,
-    'Cycle_3',  'Team_Fox',   '2020-01-01', '2020-01-14',          1,
-    'Cycle_4',  'Team_Fox',   '2020-01-15', '2020-01-30',          2
+    ~uid,       ~team_id,     ~start_date,  ~extended_end_date,    ~ordinal,  ~extended_start_date,
+    'Cycle_1',  'Team_Viper', '2020-01-01', '2020-01-14',          1,         '',
+    'Cycle_2',  'Team_Viper', '2020-01-15', '2020-01-30',          2,         '',
+    'Cycle_3',  'Team_Fox',   '2020-01-01', '2020-01-14',          1,         '',
+    'Cycle_4',  'Team_Fox',   '2020-01-15', '2020-01-30',          2,         '',
```
chris-perts, line 477: nit: this example data is surprising because Copilot is set to _always_ define `extended_start_date` for (and only for) ordinal-1 cycles. In fact that behavior is already happening in production, so we don't have to worry about old-style dates where `extended_start_date` is missing. It's fine if you're proving the RServe is actually more robust than that, but it might give readers the wrong impression.

tests/testthat/test_summarize_copilot.R
```diff
@@ -528,6 +529,89 @@ describe('map_responses_to_cycles', {
...
+      pull(cycle_ordinal)
+
+    expect_equal(p1_cycle, 1)
+
+  })
+
+  it('uses start_date_extended when start_date is absent', {
+    triton.cycle <- tribble(
+      ~uid,       ~team_id,     ~start_date,  ~extended_end_date,    ~ordinal,  ~extended_start_date,
+      'Cycle_1',  'Team_Viper', '',           '2020-01-14',          1,         '2019-06-30',
```
chris-perts, line 588: nit: this is also a little surprising, since the app's behavior is always to have both `start_date` and `end_date` set, or neither. Again, you're proving a more general case, which is fine, so this is merely a nit, but a reader might say "Oh, can that happen?"

chris-perts (approved):

R/summarize_copilot.R
sarah-gripshover-PERTS: 99c72dd

tests/testthat/test_summarize_copilot.R
sarah-gripshover-PERTS: 9707589

tests/testthat/test_summarize_copilot.R
sarah-gripshover-PERTS: 9707589

PR #85, opened 2022-01-31
Fix getting classrooms of empty networks

chris-perts:
## Background

RServe failed yesterday because there was a network with an empty set of `association_ids`. That caused an `NA` to be passed to a function which expected a string:

```
perts_ids$get_kind(): not a long uid, no underscore
```

## Implementation

I filtered `NA` values out of the network association df before continuing to work with it.

## Testing

Ran RServe locally after the fix and it got to the first report. All tests pass.


sarah-gripshover-PERTS (approved):
looks good. Thanks for debugging!

PR #87, opened 2022-07-20
Recently modified rosters

chris-perts:
## Background

Addresses https://github.com/PERTS/rserve/issues/266

Incorporated into https://github.com/PERTS/rserve/pull/268

## Implementation

To support code in the metascript, adds function `recently_modified_rosters()`. This simply looks at the modified field of the participant table, which tells us everything we need to know. Repeating the comments from that function:

```
  # The conventions of triton/Copilot allow us to detect ANY roster change by
  # looking at the `modified` field, which uses the MySQL clause
  # `ON UPDATE CURRENT_TIMESTAMP`:
  # * Adding a participant will set modified to the current time
  # * Removing a participant will _update_ the participant to be in zero
  #   classrooms, while still on the team, and thus update the modified time.
  # * Renaming a participant will update the modified time.
```
In other words because (1) participants are not truly deleted, but rather disassociated from classrooms while preserving their association with the team and (2) the database already updates the modified time on row creation or update, `modified` is all we need.

## Testing

Unit tests included.

### DIFF
--- a/tests/testthat/test_summarize_copilot.R
+++ b/tests/testthat/test_summarize_copilot.R
@@ -733,3 +733,60 @@ describe('recently_modified_cycles', {
     expect_equal(team_ids, c('Team_A'))
   })
 })
+
+describe('recently_modified_rosters', {
+  # Given these parameters, a "recently modified" entity will be one whose
+  # timestamp is on or after 2021-01-07 00:00:00.
+  report_date <- '2021-01-16'
+  time_lag_threshold <- 9
+
+  # Fields in the participant table we won't use:
+  # * participant.uid
+  # * participant.short_uid
+  # * participant.created
+  # * participant.student_id
+  # * participant.stripped_student_id
+  # * participant.classroom_ids
+  # * participant.in_target_group
+  # * participant.verified
+
+  none_modified <- dplyr::tibble(
+    participant.team_id = c('Team_A', 'Team_A', 'Team_A'),
+    participant.modified = c(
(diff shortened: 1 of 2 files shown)


tests/testthat/test_summarize_copilot.R
```diff
@@ -733,3 +733,60 @@ describe('recently_modified_cycles', {
...
+  it('includes modified participants', {
+    triton.participant <- some_modified
+    team_ids <- summarize_copilot$recently_modified_rosters(
+      triton.participant,
+      report_date,
+      time_lag_threshold
+    )
+
+    expect_equal(team_ids, c('Team_B', 'Team_C'))
+  })
```
sarah-gripshover-PERTS, line 791: nit: testing doesn't cover case where more than one modified row is present for the same team, to verify that unique teams are returned. But I can see right there in the code that you call `unique()` so whatever. If you're feeling thorough today, you might add that test. Or not. I doubt someone's going to go in there and delete your `unique()` call. (If six months from now I delete it for no reason myself after having said this, I'll buy you a beer.)

sarah-gripshover-PERTS (approved):
Looks great! Simple and to the point.

tests/testthat/test_summarize_copilot.R
chris-perts: [3145156](https://github.com/PERTS/gymnast/pull/87/commits/3145156ba8982f74b9f4ee10b5b8f53d68e40a8c)

PR #88, opened 2022-12-02
274 recent orgs

chris-perts:
Addresses https://github.com/PERTS/rserve/issues/274

9ce4402164991234d79c0aaa30837586d7ff2eea creates a function we'll need to solve the issue. Unit tests included.

f478ad2e866f89f31e83e23bc6c38c68d0c66092 is an older commit I never had QA'd. I inserted some responses into Saturn for testing dashboards in production, but they're not as well formed as they should be. Instead of fixing the responses, I just had RServe ignore them. Make me run a lap if you must, but I couldn't justify spending more than a second on appeasing RServe.

### DIFF
--- a/R/saturn.R
+++ b/R/saturn.R
@@ -123,7 +123,16 @@ create_service <- function (
   }
 
   get_responses <- function (...) {
+    qa_codes <- c(
+      'BASIC1',
+      'BASIC2',
+      'OVERLAP1',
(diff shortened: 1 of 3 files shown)


chris-perts:
Assuming @sarah-gripshover-PERTS  approval of https://github.com/PERTS/rserve/pull/275 implies this is approved also.

